PR #68, opened 2014-12-28
Introducing Action::Request

lasseebert:
Implements https://github.com/lotus/controller/issues/66

* `Action::Request` class that subclasses `Rack::Request`
* `Action#request` method that returns a `Action::Request` instance.

I have written unit tests for the methods of `Action::Request` for which I think it makes sense to guarantee backwards compatibility.

### DIFF
--- a/lib/lotus/action/rack.rb
+++ b/lib/lotus/action/rack.rb
@@ -142,6 +142,27 @@ module Lotus
         @request_id ||= SecureRandom.hex(DEFAULT_REQUEST_ID_LENGTH)
       end
 
+      # Returns a Lotus specialized rack request
+      #
+      # @return [Lotus::Action::Request] The request
+      #
+      # @since x.x.x
+      #
+      # @example
+      #   require 'lotus/controller'
+      #
+      #   class Create
+      #     include Lotus::Action
+      #
+      #     def call(params)
+      #       ip     = request.ip
+      #       secure = request.ssl?
+      #     end
+      #   end
+      def request
+        @request ||= Request.new(@_env)
+      end
+
       private
 
       # Sets the HTTP status code for the response
--- a/lib/lotus/action/request.rb
+++ b/lib/lotus/action/request.rb
@@ -1,0 +1,46 @@
+module Lotus
+  module Action
+    # A request model build on top of Rack::Request, that guarantees backwards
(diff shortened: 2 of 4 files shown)


jodosha:
@lasseebert This looks great! I'm going to merge it now, but love to hear your opinion about `#accepted_encoding` and `#accepted_language`.

Those two headers are weighted like `Accept`. For it, we have implemented a mechanism which evaluates those _"q values"_ and returns an useful information. What if we move those two informations as private methods in `Action::Rack` and let those to provide a strict result?

Eg. for a request like this `'HTTP_ACCEPT' => 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'`, `#accepts` returns `"text/html"`.

What if we do the same here? `'HTTP_ACCEPT_LANGUAGE' => 'da, en;q=0.6'`, `#language` returns `"da"` and `'HTTP_ACCEPT_ENCODING' => 'gzip, deflate;q=0.6'`, `#encoding` returns `"gzip"`.

lasseebert:
I think it's a great idea. I have never used anything but the first value in those headers. I can still imagine situations where it is useful to know the prioritized list of values (but without the q values):

* `language # => The top prioritzed language`
* `languages # => Prioritized list of languages, e.g. ['da', 'en']`
* `encoding # => The top prioritzed encoding`
* `encodings # => Prioritized list of encodings, e.g. ['gzip', 'deflate']`

jodosha:
@lasseebert I'd say to keep it consistent with MIME types logic: return only one value and expose a query method to understand if a value is accepted.

```ruby
require 'lotus/controller'

class Show
  include Lotus::Action

  def call(params)
    # ...
    # @_env['HTTP_ACCEPT'] # => "text/html,application/xhtml+xml,application/xml;q=0.9"

    accepts # => "text/html"

    accept?('text/html')        # => true
    accept?('application/xml')  # => true
    accept?('application/json') # => false
  end
end
```

So we could have `#language`, `#accept_language?` and `#encoding`, `#accept_encoding?` pairs.

lasseebert:
This makes sense if there is still access to the methods on `Action::Request`: `#accept_language` and `#accept_encoding`.

Otherwise we would hide the logic that parses the accept lists, and people who need it will probably create a `Rack::Request`, which was the point to avoid.

PR #69, opened 2015-01-07
Fix param coercion with Lotus Validations

stevehodgkiss:
When using whitelisted parameters, `params[:whatever]` is always nil. Also, type coercion was broken.

I'm not too pleased with the conditionals though, perhaps we could use a separate class for `WhitelistedParams` and `Params`?

Also, there's one failing test I'm unsure of... and I'm heading out now.

### DIFF
--- a/test/action/params_test.rb
+++ b/test/action/params_test.rb
@@ -128,8 +128,9 @@ describe Lotus::Action::Params do
     before do
       TestParams = Class.new(Lotus::Action::Params) do
         param :email, presence:   true, format: /\A.+@.+\z/
-        param :name,  presence:   true
+        param :name,  type: String, presence:   true
         param :tos,   acceptance: true
+        param :age,   type: Integer
       end
     end
 
@@ -152,6 +153,12 @@ describe Lotus::Action::Params do
 
       params.valid?.must_equal true
       params.errors.must_be_empty
+    end
+
+    it "coerces input" do
+      params = TestParams.new(name: 'John', age: '1')
+      params.name.must_equal('John')
+      params.age.must_equal(1)
     end
   end
 
--- a/test/action/params_test.rb
+++ b/test/action/params_test.rb
@@ -155,10 +155,11 @@ describe Lotus::Action::Params do
       params.errors.must_be_empty
     end
 
-    it "coerces input" do
+    it "has input available as methods" do
       params = TestParams.new(name: 'John', age: '1')
       params.name.must_equal('John')
       params.age.must_equal(1)
+      params[:age].must_equal(1)
     end
   end
 
--- a/lib/lotus/action/params.rb
+++ b/lib/lotus/action/params.rb
@@ -118,7 +118,11 @@ module Lotus
       #
       # @since 0.2.0
       def [](key)
-        @attributes.get(key)
+        if self.class.whitelisting?
+          public_send(key) if self.class.defined_attributes.include?(key.to_s)
+        else
+          @attributes.get(key)
+        end
       end
 
       # Returns the Ruby's hash
@@ -127,7 +131,11 @@ module Lotus
       #
       # @since 0.3.0
       def to_h
-        @attributes.to_h
+        if self.class.whitelisting?
+          Utils::Hash.new(super).stringify!
+        else
+          @attributes.to_h
+        end
       end
       alias_method :to_hash, :to_h
 
@@ -137,7 +145,11 @@ module Lotus
       # @since 0.3.1
       # @api private
       def read_attributes
-        to_h
+        if self.class.whitelisting?
+          super
+        else
+          to_h
+        end
       end
 
       # @since 0.3.1
--- a/lib/lotus/action/params.rb
+++ b/lib/lotus/action/params.rb
@@ -106,7 +106,7 @@ module Lotus
       # @since 0.1.0
       def initialize(env)
         @env = env
-        _compute_params
+        super(_compute_params)
         # freeze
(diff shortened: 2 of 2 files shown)


test/action/params_test.rb
```diff
@@ -153,6 +154,13 @@
       params.valid?.must_equal true
       params.errors.must_be_empty
     end
+
+    it "has input available as methods" do
```
jodosha: This is a nice to have feature, but we should get a subscriber operator (`#[]`) correct too. ;)

jodosha:
@stevehodgkiss The CI failure is due to those mismatching values:

```
# ...
:errors=>{:__errors=>[#<Lotus::Validations::Error:0x0000000327a940 @attribute=:title, @validation=:presence, @expected=true, @actual=nil>]}
```

Notice the bit `@actual=nil`. We pass a blank string in [the test](https://github.com/lotus/controller/blob/master/test/integration/full_stack_test.rb#L20), and we [assert](https://github.com/lotus/controller/blob/master/test/integration/full_stack_test.rb#L24) that the actual value is a blank string.

jodosha:
@stevehodgkiss I've tried to fiddle with subclasses and modules.

### Approach 1: Instance extend

Init and then do `self.extend(WhitelistedParams)`. This works but confuses Ruby VM method caching, slowing down the object. Read more.

### Approach 2: Subclassing

The right place where to hook this functionality is `Validatable.params`, where we create an anonymous class which uses `Params` as superclass. I've created `WhitelistedParams` which inherit from `Params`.

The problem is duplication. Where you have introduced the conditional ( `if whitelisting?` ) you call `super` to hook into `Lotus::Validations` which is a direct parent for `Params`.

When subclassing, call `super` is no longer possible. That means to manually duplicate the code from Validations. Which is overkilling for maintenance.

### Approach 3: Extending

I've transformed `WhitelistedParams` into a module with the goal of prepending it and be able to call `super` again. The hook point is always the same: `Validatable.params`.

```ruby
        def params(klass = nil, &blk)
          if block_given?
            klass = Class.new(Params)
            klass.prepend(WhitelistedParams)
            klass.class_eval(&blk)

            @params_class = const_set(PARAMS_CLASS_NAME, klass)
          else
            @params_class = klass
          end
        end
```

But it causes tons of errors like:

```
1337) Error:
Lotus::Action::.handle_exception#test_0001_handle an exception with the given status:
NoMethodError: undefined method `errors' for nil:NilClass
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/validatable.rb:130:in `errors'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/exposable.rb:98:in `block (2 levels) in exposures'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/exposable.rb:97:in `each'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/exposable.rb:97:in `block in exposures'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/exposable.rb:96:in `tap'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/exposable.rb:96:in `exposures'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/exposable.rb:111:in `finish'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/callable.rb:89:in `finish'
    /Users/luca/Code/stevehodgkiss-controller/lib/lotus/action/callable.rb:76:in `call'
    /Users/luca/Code/stevehodgkiss-controller/test/throw_test.rb:10:in `block (3 levels) in <top (required)>'
```

### What to do

I think that validations should work with a set of `@attributes`.
It simplifies the life for us when dealing with Params.

Do you think it's possible to do so?

What if we change the code of `Validations::AttributeDefiner` like this?

```ruby
        def define_coerced_writer(name, type)
          class_eval %{
            def #{ name }=(value)
              @attributes.set(name.to_s, Lotus::Validations::Coercions.coerce(#{ type.name }, value))
            end
          }
        end

        # new method
       def define_reader(name)
           class_eval %{
            def #{ name }
              @attributes.get(name.to_s)
            end
          }
       end
```

At the same time:

```ruby
      def initialize(attributes)
        @attributes = Utils::Attributes.new
        attributes.to_h.each do |key, value|
          writer = "#{ key }="
          public_send(writer, value) if respond_to?(writer)
        end
      end
```

### Bottom line

I agree with you: this shouldn't be a Validations' business. But it's paradoxical that the new changes creates extensions points for unknown libraries, but we aren't able to support the only well known library that needs it.

Instance variables are much faster than `Utils::Attributes`, but this is the price to pay for correctness and security. The first value is about coercions and the second is about non-collectable symbols.

Make it correct, then make it faster.

/cc @joneslee85

stevehodgkiss:
@jodosha validations PR for the CI failure issue (related to coercions) https://github.com/lotus/validations/pull/40

stevehodgkiss:
@jodosha agreed, if it makes sense for this to live in validations and the code is simpler let's work towards that. We just need to make sure we read attributes from their accessors for validation purposes (so other object builders can be used).

test/action/params_test.rb
```diff
@@ -128,8 +128,9 @@
     before do
       TestParams = Class.new(Lotus::Action::Params) do
         param :email, presence:   true, format: /\A.+@.+\z/
```
joneslee85: _nitpick_, replace 2 spaces with 1 spaces after `presence:`

PR #73, opened 2015-01-16
#halt now accepts an optional message argument

weppos:
When #message is given, it sets the response body with the message.

This is useful in case you need to set a different response for the same HTTP error. For instance, a 403 may be caused by an invalid password, an invalid username, or a bad OAUTH token.

### DIFF
--- a/lib/lotus/action/throwable.rb
+++ b/lib/lotus/action/throwable.rb
@@ -62,24 +62,29 @@ module Lotus
 
       protected
 
-      # Halt the action execution with the given HTTP status code.
+      # Halt the action execution with the given HTTP status code and message.
       #
       # When used, the execution of a callback or of an action is interrupted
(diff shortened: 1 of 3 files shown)


weppos:
Real world example, given the following error helper

```ruby
def error(code, message)
  self.format = :json
  halt(code, { message: message }.to_json)
end
```

I'd like to be able to perform

```ruby
def authenticate!
  authentication_context or error(401, I18n.t("api.message_error_authentication_failed"))
end
```

or to use the helper as a result of an exception handler.

```
def handle_record_not_found(exception)
  halt(404, { message: exception.message }.to_json)
end
```

AlfonsoUceda:
:+1: looks good

jodosha:
@weppos Thanks for this PR it looks good, merged :+1:

PR #77, opened 2015-01-20
Fix rack.errors management

### LINKED ISSUE #74: Lotus::Action should not dump the exception if the exception is handled
Given the following action

```ruby
class Show
  include Api::V2::Action

  handle_exception RecordNotFound => :handle_record_not_found

  def call(params)
    # ...
  end

  def handle_record_not_found(exception)
    halt(404, { message: exception.message }.to_json)
  end
end
```

An HTTP request for an invalid resource will correctly render the error message

```
{"message":"Domain `3999' not found"}
```

However, the exception is also dumped in the log

```
ActiveRecord::RecordNotFound: Domain `3999' not found
  ...
  ...
  ...
  /Users/weppos/.rvm/gems/ruby-2.1.5@dnsimple/gems/lotus-controller-0.3.1/lib/lotus/action/callbacks.rb:149:in `call'
  /Users/weppos/.rvm/gems/ruby-2.1.5@dnsimple/gems/lotus-controller-0.3.1/lib/lotus/action/callable.rb:73:in `block in call'
  /Users/weppos/.rvm/gems/ruby-2.1.5@dnsimple/gems/lotus-controller-0.3.1/lib/lotus/action/throwable.rb:104:in `block in _rescue'
  /Users/weppos/.rvm/gems/ruby-2.1.5@dnsimple/gems/lotus-controller-0.3.1/lib/lotus/action/throwable.rb:102:in `catch'
  /Users/weppos/.rvm/gems/ruby-2.1.5@dnsimple/gems/lotus-controller-0.3.1/lib/lotus/action/throwable.rb:102:in `_rescue'
```

making it hard to distinguish handled vs not-handled exceptions. That's because `_rescue` always dumps the error, no matter if it was handled or not.

```ruby
      def _rescue
        catch :halt do
          begin
            yield
          rescue => exception
            _reference_in_rack_errors(exception)
            _handle_exception(exception)
          end
        end
      end
```

My proposal is to dump the exception only if not handled. Though?

jodosha:
Only dump exceptions in rack.errors if handling is turned off, or the raised exception isn't managed

Closes #74 

/cc @joneslee85 and @weppos

### DIFF
--- a/lib/lotus/action/throwable.rb
+++ b/lib/lotus/action/throwable.rb
@@ -138,6 +138,8 @@ module Lotus
       # @since 0.2.0
       # @api private
       def _reference_in_rack_errors(exception)
+        return if configuration.handled_exception?(exception)
+
         if errors = @_env[RACK_ERRORS]
           errors.write(_dump_exception(exception))
           errors.flush
--- a/lib/lotus/controller/configuration.rb
+++ b/lib/lotus/controller/configuration.rb
@@ -194,6 +194,19 @@ module Lotus
         @handled_exceptions.fetch(exception.class) { DEFAULT_ERROR_CODE }
       end
 
+      # Check if the given exception is handled.
+      #
+      # @param exception [Exception] an exception
+      #
+      # @since x.x.x
+      # @api private
+      #
+      # @see Lotus::Controller::Configuration#handle_exception
+      def handled_exception?(exception)
+        handled_exceptions &&
+          !!@handled_exceptions.fetch(exception.class) { false }
+      end
+
       # Specify which is the default action module to be included when we use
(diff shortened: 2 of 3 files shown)


weppos:
:+1: PS: you could have simply changed the order of the methods from

```ruby
rescue => exception
  _reference_in_rack_errors(exception)
  _handle_exception(exception)
end
```

to

```ruby
rescue => exception
  _handle_exception(exception) || _reference_in_rack_errors(exception)
  
end
```

Too magic: :wink:

joneslee85:
:+1:

jodosha:
@weppos Three specs fail and yes, it's too magic ;)

Issue #72, opened 2015-01-15
Rack middleware is not working correctly


benedikt:
The current implementation of `Lotus::Action::Rack.use` doesn't support passing the app to the middleware. To allow intercepting of requests and responses, Rack middleware is usually implemented like this:

```ruby
class Middleware
  def initialize(app)
    @app = app
  end

  def call(env)
    @app.call(env)
  end
end
```

The current middleware implementation just passes the `env` to the middleware's `call` method and therefore doesn't work with middleware that might stop the execution of the middleware stack or modify the request or the response.

A reference implementation of Rack middleware can be found in Rack itself:

1. https://github.com/rack/rack/blob/master/lib/rack/builder.rb#L81-L87
1. https://github.com/rack/rack/blob/master/lib/rack/builder.rb#L144-L154

Thanks a lot!

jodosha:
@benedikt Hello and thanks for reporting this.
I do agree with you: currently we only have an half-backed solution for fine grained rack stacks.

We'll try to accommodate this change soon. Which middleware are you trying to mount?

benedikt:
Hi @jodosha,

thanks for looking into this! At the moment, I'm mostly looking at authentication middleware like warden as well as custom middleware.
I'm happy to help out, but I'm not very familiar with the overall architecture and ideas of this gem, yet.

jodosha:
@benedikt Hey, thanks for the quick reply. Mr @AlfonsoUceda will be in charge of this. We're discussing about details right now.

jodosha:
@benedikt me and @AlfonsoUceda looked at this on yesterday.

We have found a solution, but it still needs to be validated with code.

benedikt:
Nice! Take your time. It's not super urgent :)

AlfonsoUceda:
@benedikt fixed in master in lotus/controller and lotus/router, this will be released soon ;)

jodosha:
@benedikt can you please try by using the following `Gemfile` configuration? Please report if it works or if there are other issues. We want to make sure that this works for you before to release it. Thank you. :hamster: 

```ruby
gem 'lotus-router', github: 'lotus/router'
gem 'lotus-controller', github: 'lotus/controller'
```

jodosha:
I've tested this feature in isolation here: https://gist.github.com/jodosha/685b6369f9aeb1b01bf7

Closing this. @benedikt Feel free to reopen if it doesn't work for you.

benedikt:
[...]

PR #78, opened 2015-01-21
Refactorized action rack implementation

AlfonsoUceda:
reference #72 

Be careful before merging because the lotus-router is pointing to my fork and an especific branch related with this lotus/router#36 (later this will be restored to lotus/router github)

Thanks you @jodosha for the work

### DIFF
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -170,19 +170,55 @@ class ExposeAction
   end
 end
 
-class XMiddleware
-  def self.call(env)
-    env['X-Middleware'] = 'OK'
+class YMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    code, headers, body = @app.call(env)
+    [code, headers.merge!('Y-Middleware' => 'OK'), body]
   end
 end
 
-class UseAction
-  include Lotus::Action
+class XMiddleware
+  def initialize(app)
+    @app = app
+  end
 
-  use XMiddleware
+  def call(env)
+    code, headers, body = @app.call(env)
+    [code, headers.merge!('X-Middleware' => 'OK'), body]
+  end
+end
 
-  def call(params)
-    headers['X-Middleware'] = params.env.fetch('X-Middleware')
+module UseAction
+  class Index
+    include Lotus::Action
+    use XMiddleware
+
+    def call(params)
+      self.body = 'Hello from UseAction::Index'
+    end
+  end
+
+  class Show
+    include Lotus::Action
+    use YMiddleware
+
+    def call(params)
+      self.body = 'Hello from UseAction::Show'
+    end
+  end
+end
+
+module NoUseAction
+  class Index
+    include Lotus::Action
+
+    def call(params)
+      self.body = 'Hello from NoUseAction::Index'
+    end
   end
 end
 
--- a/test/integration/use_test.rb
+++ b/test/integration/use_test.rb
@@ -1,13 +1,59 @@
 require 'test_helper'
+require 'rack/test'
 
 describe 'Rack middleware integration' do
-  describe '.use' do
-    it 'uses the specified Rack middleware' do
-      response = Rack::MockRequest.new(UseAction.new).get('/')
-      status, headers, _ = response
+  include Rack::Test::Methods
 
-      status.must_equal 200
-      headers['X-Middleware'].must_equal 'OK'
+  def response
+    last_response
+  end
+
+  describe '.use' do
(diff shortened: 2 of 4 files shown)


joneslee85:
Looks good :+1:

AlfonsoUceda:
I need to create want test for rack builder in a couple of hours. Yesterday I forgot it

test/integration/use_test.rb
```diff
@@ -1,13 +1,69 @@
...
+        get '/show', to: 'use_action#show'
+      end
+
+      DoubleUseActionApplication = Rack::Builder.new do
+        run router
+      end.to_app
+
+      get '/'
+
+      UseAction::Index.rack_builder == UseAction::Show.rack_builder
```
AlfonsoUceda: is this correct?

```ruby
UseAction::Index.rack_builder == UseAction::Show.rack_builder # => true
```
```ruby
UseAction::Index.rack_builder.must_equal UseAction::Show.rack_builder
# => Fail
# No visible difference in the Rack::Builder#inspect output.
# You should look at the implementation of #== on Rack::Builder or its members.
```
```ruby
UseAction::Index.rack_builder.object_id.must_equal UseAction::Show.rack_builder.object_id # => false
```

jodosha: I don't think this would add a value. We're testing a detail, instead of a behavior. Also, if Rack Team customizes `Rack::Builder#==` this test will fail in the future.

The behavior that we want to describe is: given two actions with separated middleware stacks, those shall not to interfers each other.

What if we introduce `YMiddleware` and let `UseAction::Show` to use it? Then we could introspect the response and headers returned from `/show`.

AlfonsoUceda: ok, then I'll rewrite the test ;)

test/integration/use_test.rb
```diff
@@ -1,13 +1,59 @@
...
+        get '/show', to: 'use_action#show'
+      end
+
+      UseActionApplication = Rack::Builder.new do
+        run router
+      end.to_app
+
+      get '/'
+
+      UseAction::Index.respond_to?(:call).must_equal true
```
jodosha: This is a low level detail that we can safely remove from here. The test already covers it. :cat:

test/integration/use_test.rb
```diff
@@ -1,13 +1,59 @@
...
+      response.status.must_equal 200
+      response.headers.fetch('X-Middleware').must_equal 'OK'
+      response.headers['Y-Middleware'].must_be_nil
+
+      get '/show'
+
+      UseAction::Show.respond_to?(:call).must_equal true
```
jodosha: Likewise: https://github.com/lotus/controller/pull/78/files#r23456293 :mouse:

test/fixtures.rb
```diff
@@ -170,19 +170,52 @@ def call(params)
...
 end
 
-class UseAction
-  include Lotus::Action
+module UseAction
+  class Index
+    include Lotus::Action
+    use XMiddleware
+
+    def call(params)
```
jodosha: What do you think if we add the following line and prove it from the test?

```ruby
def call(params)
  self.body = 'foo'
end
```

.. and ..

```ruby
response.body.must_equal ['foo']
```

:beetle:

test/fixtures.rb
```diff
@@ -170,19 +170,52 @@ def call(params)
...
+  end
 
-  use XMiddleware
+  class Show
+    include Lotus::Action
+    use YMiddleware
 
-  def call(params)
-    headers['X-Middleware'] = params.env.fetch('X-Middleware')
+    def call(params)
```
jodosha: Likewise: https://github.com/lotus/controller/pull/78/files#r23456473

:dancers:

test/integration/use_test.rb
```diff
@@ -1,13 +1,59 @@
...
+        run router
+      end.to_app
+
+      get '/'
 
-      status.must_equal 200
-      headers['X-Middleware'].must_equal 'OK'
+      NoUseAction::Index.respond_to?(:call).must_equal false
+      response.status.must_equal 200
+      -> { response.headers.fetch('X-Middleware') }.must_raise KeyError
```
jodosha:
```ruby
response.headers['X-Middleware'].must_be_nil
```

:cake:

jodosha:
@AlfonsoUceda This looks good! :sparkles: :+1: for me.

AlfonsoUceda:
@jodosha thanks for your help ;)

joneslee85:
@AlfonsoUceda :sparkle: :+1: great work

PR #82, opened 2015-01-27
Added raw attribute to params to get all request's params

### LINKED ISSUE #79: Ability to access unfiltered request parameters
This is the result of a chat I had with @jodosha about a real-world scenario.

## Scenario

Let's suppose I configure an action with some param whitelisting.

```ruby
class Signup
  include Lotus::Action

  params do
    param :first_name
    param :last_name
    param :email
  end

  def call(params)
    # ...
  end
end
```

Once the action is triggered, the `params` object will return a filtered list of params. Therefore, if the action is called with

    POST /signup?first_name=Simone&last_name=Carletti&email=whatever&foo=bar

the action will strip the unallowed `foo` parameter. In most cases, this is what we want. However, let's assume that I have a command (similar to the Interactor @jodosha's idea) that I use to process the signup, that takes as input a whitelisted Hash of signup attributes.

```ruby
class Signup
  include Lotus::Action

  params do
    param :first_name
    param :last_name
    param :email
  end

  def call(params)
    SignupInteractor.execute(params)
  end
end
```

So far so good. But our use case is a little bit more complex. We want to allow people to signup, and being redirected to a specific `return_to` path passed in query string, or a default url if the return path is not present.

```ruby
class Signup
  include Lotus::Action

  params do
    param :first_name
    param :last_name
    param :email
  end

  def call(params)
    SignupInteractor.execute(params)

    redirect_to params[:return_to] ? params[:return_to] : 'default'
  end
end
```

Here we have a problem. I don't have access anymore to the `return_to`. There are a number of possible workarounds, however both @jodosha and I agree that there should be a way to access the unfiltered list of params in a request.

This is just a possible use case, but there are a lot of them.

## API

This is a call for feedback. Some options are:

- `params.raw[:whatever]`
- `params.original[:whatever]`

Proposals?

AlfonsoUceda:
Closes #79

### DIFF
--- a/lib/lotus/action/params.rb
+++ b/lib/lotus/action/params.rb
@@ -97,6 +97,11 @@ module Lotus
       # @api private
       attr_reader :env
 
+      # @attr_reader raw [Lotus::Utils::Attributes] all request's attributes
+      #
+      # @since x.x.x
+      attr_reader :raw
+
       # Initialize the params and freeze them.
       #
       # @param env [Hash] a Rack env or an hash of params.
@@ -138,13 +143,17 @@ module Lotus
         if self.class.whitelisting?
           _whitelisted_params
         else
-          @attributes = Utils::Attributes.new(_params)
+          @attributes = _raw
         end
       end
 
       # @since 0.3.1
       # @api private
-      def _params
+      def _raw
+        @raw ||= Utils::Attributes.new(_create_raw)
+      end
+
+      def _create_raw
         {}.tap do |result|
           if env.has_key?(RACK_INPUT)
             result.merge! ::Rack::Request.new(env).params
@@ -159,7 +168,7 @@ module Lotus
       # @api private
       def _whitelisted_params
         {}.tap do |result|
-          _params.each do |k, v|
+          _raw.to_h.each do |k, v|
             next unless self.class.defined_attributes.include?(k.to_s)
 
             result[k] = v
--- a/test/action/params_test.rb
+++ b/test/action/params_test.rb
@@ -12,6 +12,36 @@ describe Lotus::Action::Params do
   #   params.must_be :frozen?
   # end
 
+  describe 'raw params' do
+    before do
+      @params = Class.new(Lotus::Action::Params)
+    end
+
+    describe "when this feature isn't enabled" do
+      before do
+        @action = ParamsAction.new
+      end
+
+      it 'raw gets all params' do
+        @action.call({id: 1, unknown: 2})
+        @action.params.raw.get(:id).must_equal 1
+        @action.params.raw.get(:unknown).must_equal 2
+      end
+    end
+
+    describe "when this feature is enabled" do
+      before do
+        @action = WhitelistedParamsAction.new
+      end
+
+      it 'raw gets all params' do
+        @action.call({id: 1, unknown: 2})
+        @action.params.raw.get(:id).must_equal 1
+        @action.params.raw.get(:unknown).must_equal 2
+      end
+    end
+  end
+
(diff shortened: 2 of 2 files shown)


jodosha:
@AlfonsoUceda Thanks for this PR!

I'm wondering if it's worth to keep both `@raw` and `@attributes` around. If they have both the same value can we please consider to have just one? Thank you!

AlfonsoUceda:
I saw one thing with these code:
```ruby
2.1.5 :001 > 
2.1.5 :002 >   require 'lotus/controller'
 => true 
2.1.5 :003 > 
2.1.5 :004 >   class WhitelistedParamsAction
2.1.5 :005?>     include Lotus::Action
2.1.5 :006?>   
2.1.5 :007 >       params do
2.1.5 :008 >           param :id
2.1.5 :009?>       end
2.1.5 :010?>   
2.1.5 :011 >       def call(params)
2.1.5 :012?>         puts params
2.1.5 :013?>         self.body = params.to_h.inspect
2.1.5 :014?>       end
2.1.5 :015?>   end
 => :call 
2.1.5 :016 > 
2.1.5 :017 >   class ParamsAction
2.1.5 :018?>     include Lotus::Action
2.1.5 :019?>   
2.1.5 :020 >       def call(params)
2.1.5 :021?>         puts params
2.1.5 :022?>         self.body = params.to_h.inspect
2.1.5 :023?>       end
2.1.5 :024?>   end
 => :call 
2.1.5 :025 > 
2.1.5 :026 >   action = WhitelistedParamsAction.new
 => #<WhitelistedParamsAction:0x00000001513bb8> 
2.1.5 :027 > action.call({id: 1, foo: 2})
#<WhitelistedParamsAction::Params:0x000000015028b8>
 => [200, {"Content-Type"=>"application/octet-stream; charset=utf-8"}, ["{\"id\"=>1}"]] 
2.1.5 :028 > action.params
 => #<WhitelistedParamsAction::Params:0x000000015028b8 @env={:id=>1, :foo=>2}, @raw=#<Lotus::Utils::Attributes:0x00000001502520 @attributes={"id"=>1, "foo"=>2}>, @attributes=#<Lotus::Utils::Attributes:0x00000001501f58 @attributes={"id"=>1}>, @errors=#<Lotus::Validations::Errors:0x000000015010d0 @errors={}>> 
2.1.5 :029 > action = ParamsAction.new
 => #<ParamsAction:0x0000000262f4b0> 
2.1.5 :030 > action.call({id: 1, foo: 2})
#<ParamsAction::Params:0x00000002631c88>
 => [200, {"Content-Type"=>"application/octet-stream; charset=utf-8"}, ["{\"id\"=>1, \"foo\"=>2}"]] 
2.1.5 :031 > action.params
 => #<ParamsAction::Params:0x00000002631c88 @env={:id=>1, :foo=>2}, @raw=#<Lotus::Utils::Attributes:0x00000002631b98 @attributes={"id"=>1, "foo"=>2}>, @attributes=#<Lotus::Utils::Attributes:0x00000002631b98 @attributes={"id"=>1, "foo"=>2}>, @errors=#<Lotus::Validations::Errors:0x000000026316c0 @errors={}>>
```
If you use the params method inside action, the action's params' @attributes are `id` and `foo`, but if you use the params method  the action's params' @attributes are the whitelisted params `id`

AlfonsoUceda:
@jodosha `@attributes` comes from https://github.com/lotus/validations/blob/master/lib/lotus/validations/attribute_definer.rb

jodosha:
@AlfonsoUceda This looks great! :sparkles: thank you! :+1:

PR #83, opened 2015-01-29
Added new callbacks to action.

### LINKED ISSUE #76: Ability to configure a filter in a specific position
Granted that one of @jodosha's goal was to keep callbacks as simple as possible, I have a case that I feel reveal a reasonable improvement.

Let's assume I have a filter called `authenticate!` that must be run on every action.

```ruby
class Show
  before :authenticate!

  def call(params)
  end
end
```

On a **very few actions** (that's one of the key conditions), let's say 10 out of 100, I need a specific validation to happen before the authenticate.

```ruby
class Index
  before :validate!, :authenticate!

  def call(params)
  end
end
```

I can indeed stick with the explicit example above, but that would mean I can't extract the `before :authenticate!` into a controller config, or a module.

Therefore, I wonder what do you think about providing a finer way to control where I want to add my filter. Right now, there's only a `Chain#add` method.

I propose to add something similar to

- `Chain#append(callback)`
- `Chain#prepend(callback)`
- `Chain#insert(before, callback)` (or `Chain#insert_before(before, callback)` and `Chain#insert_after(after, callback)`)

I can work on a patch, if we agree on naming and API.

AlfonsoUceda:
Added prepend_before, prepend_after, append_before and append_after.

Closes #76

### DIFF
--- a/lib/lotus/action/callbacks.rb
+++ b/lib/lotus/action/callbacks.rb
@@ -65,6 +65,8 @@ module Lotus
         #
         # @since 0.1.0
         #
+        # @see Lotus::Action::Callbacks::ClassMethods#append_after
+        #
         # @example Method names (symbols)
         #   require 'lotus/controller'
         #
@@ -111,9 +113,11 @@ module Lotus
         #   # 1. authentication
         #   # 2. set the article
         #   # 3. #call
-        def before(*callbacks, &blk)
+        def append_before(*callbacks, &blk)
           before_callbacks.append(*callbacks, &blk)
         end
+
+        alias_method :before, :append_before
 
         # Define a callback for an Action.
         # The callback will be executed **after** the action is called, in the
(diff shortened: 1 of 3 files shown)


jodosha:
@AlfonsoUceda This looks great for me. :+1:

AlfonsoUceda:
I didn't introduce more tests, I think is enough

jodosha:
@joneslee85 and @weppos Do you mind to review this?

weppos:
One concern. I wonder if the primary method should be `append_before` and the alias `before`, rather than the opposite, for consistency with `prepend_before`.

jodosha:
@weppos That was mine too. Now we're two people concerned with it. :+1: for this change.
If there isn't any stopping point, this should be released on tomorrow as `v0.3.2`.

AlfonsoUceda:
ok @jodosha and @weppos I'm going to do that change ;) :+1: for the change

joneslee85:
me like, me :+1:

jodosha:
@AlfonsoUceda I've merged this, thank you! :+1:

weppos:
Thanks @AlfonsoUceda for the work. :clap:

AlfonsoUceda:
you welcome mates ;)

PR #85, opened 2015-02-04
Simplify .travis.yml as jruby-head is in 2.2 mode by default

PragTob:
* old JRuby is old, no need to test

FYI I also filed a bug for the one failing spec on JRuby here - yay open source :)

Cheers,
Tobi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,8 @@ rvm:
   - 2.1.5
   - 2.2
   - rbx-2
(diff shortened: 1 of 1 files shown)


joneslee85:
@PragTob :heart:

PragTob:
Oh yeah here it goes: jruby/jruby#2563

Shouldn't hold up this PR, though :)

PR #88, opened 2015-02-13
Added default headers option

### LINKED ISSUE #87: Default headers
Allow the configuration of default headers. All the response will send those default values.
Developers MUST have the chance to override one or more headers in each single action.

Default headers are expressed as a Hash, where keys are the header name and values the header value. The default value SHOULD be an empty Hash.

```ruby
  default_headers({
    'X-Foo' => 'bar',
    'Content-Security-Policy' => '...'
  })
end
```

AlfonsoUceda:
closes  #87

### DIFF
--- a/lib/lotus/action/head.rb
+++ b/lib/lotus/action/head.rb
@@ -14,6 +14,25 @@ module Lotus
       # @api private
       HTTP_STATUSES_WITHOUT_BODY = Set.new((100..199).to_a << 204 << 205 << 304).freeze
 
+
+      # Entity headers that by RFC are permitted
+      #
+      # @since x.x.x
+      # @api private
+      ENTITY_HEADERS = {
+        'Allow'            => true,
+        'Content-Encoding' => true,
+        'Content-Language' => true,
+        'Content-Length'   => true,
+        'Content-Location' => true,
+        'Content-MD5'      => true,
+        'Content-Range'    => true,
+        'Content-Type'     => true,
+        'Expires'          => true,
+        'Last-Modified'    => true,
+        'extension-header' => true
+      }.freeze
+
       # Ensures to not send body or headers for HEAD requests and/or for status
       # codes that doesn't allow them.
       #
@@ -26,6 +45,7 @@ module Lotus
 
         if _requires_no_body?
           @_body = nil
+          @headers.reject! { |header,_| !ENTITY_HEADERS.include?(header) }
         end
       end
 
--- a/lib/lotus/controller/configuration.rb
+++ b/lib/lotus/controller/configuration.rb
@@ -469,6 +469,33 @@ module Lotus
         end
       end
 
+      # Set default headers for all responses
+      #
+      # By default this value is an empty hash.
+      #
+      # @since x.x.x
+      #
+      # @example Getting the value
+      #   require 'lotus/controller'
+      #
+      #   Lotus::Controller.configuration.default_headers # => {}
+      #
+      # @example Setting the value
+      #   require 'lotus/controller'
+      #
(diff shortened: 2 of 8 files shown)


lib/lotus/action/head.rb
```diff
@@ -26,6 +26,8 @@ def finish
 
         if _requires_no_body?
           @_body = nil
+        else
+          @headers.merge!(configuration.default_headers)
```
AlfonsoUceda: I added this else for merging the default headers if the method isn't HEAD or the status code isn't HTTP_STATUSES_WITHOUT_BODY defined above

test/configuration_test.rb
```diff
@@ -270,6 +270,26 @@ def hash
...
+        @configuration.default_headers(headers)
+      end
+
+      it 'returns the value' do
+        @configuration.default_headers.must_equal headers
+      end
+    end
+  end
+
   describe 'duplicate' do
```
jodosha: Can you please add this to `duplicate` and `reset!` tests? TY!

jodosha:
@AlfonsoUceda This looks really good. I left one comment for extra tests for `Configuration#duplicate` and `#reset`.

Quoting #87 :

> Developers MUST have the chance to override one or more headers in each single action.

This is still missing and it's really important for resource based headers.

Can you please ensure that it works with a test like this?

```ruby
class MyAction
  include Lotus::Action

  # Default header in configuration is "Content-Security-Policy" => "script-src 'self'"
  def call(params)
    # ...
    headers['Content-Security-Policy'] = "script-src 'self' https://api.example.org"
  end
end
```

The header specified in the action should be the output.

lib/lotus/action/head.rb
```diff
@@ -26,6 +45,7 @@ def finish
 
         if _requires_no_body?
           @_body = nil
+          @headers.reject! {|header,_| !ENTITY_HEADERS.include?(header) }
```
joneslee85: nitpick, add a space after `{`

lib/lotus/action/head.rb
```diff
@@ -14,6 +14,25 @@ module Head
...
+        'Content-Encoding' => true,
+        'Content-Language' => true,
+        'Content-Length'   => true,
+        'Content-Location' => true,
+        'Content-MD5'      => true,
+        'Content-Range'    => true,
+        'Content-Type'     => true,
+        'Expires'          => true,
+        'Last-Modified'    => true,
+        'extension-header' => true
```
joneslee85: is it supposed to be all lowered case?

lib/lotus/controller/configuration.rb
```diff
@@ -469,6 +469,33 @@ def default_charset(charset = nil)
         end
       end
 
+      # Set default headers for all the responses
```
joneslee85: /s/the//

test/integration/full_stack_test.rb
```diff
@@ -20,7 +20,7 @@ def app
     head '/head', {}, 'HTTP_ACCEPT' => 'text/html'
 
     last_response.body.must_be_empty
-    last_response.headers['X-Renderable'].must_equal 'false'
+    last_response.headers['X-Renderable'].must_equal nil
```
joneslee85: must_be_nil instead?

lib/lotus/action/head.rb
AlfonsoUceda: I don't know, I'll take from here https://github.com/rack/rack/issues/787

joneslee85:
except above comments, :+1: to go

jodosha:
@AlfonsoUceda Thank you, merged :wink:

PR #89, opened 2015-02-20
Introduced Lotus::Action::Params#get

jodosha:
## What

Introduced `Lotus::Action::Params#get` as a safe way to access (nested) attributes.

## Why?

  1. Developers can **safely access nested params** without worrying of `NoMethodErrors` that can be caused by the subscriber notation. Eg `params[:address][:street]`, if `:address` isn't sent, the action will blow up. The lack of this new suggested API (`#get`) will force developers to write defensive code to access nested attributes.
  2. This serves as supporting feature for Lotus::Helpers' form helper. In case of param validation failures, a form should be able to be rendered again and **keep the current values**.

## Usage

```ruby
require 'lotus/controller'

module Deliveries
  class Create
    include Lotus::Action

    params do
      param :customer_name
      param :address do
        param :city
      end
    end

    def call(params)
      params.get('customer_name')   # => "Luca"
      params.get('uknown')          # => nil

      params.get('address.city')    # => "Rome"
      params.get('address.unknown') # => nil

      params.get(nil)               # => nil
    end
  end
end
```

### DIFF
--- a/lib/lotus/action/params.rb
+++ b/lib/lotus/action/params.rb
@@ -26,6 +26,14 @@ module Lotus
       # @since 0.1.0
       ROUTER_PARAMS = 'router.params'.freeze
 
+      # Separator for #get
+      #
+      # @since x.x.x
+      # @api private
+      #
+      # @see Lotus::Action::Params#get
+      GET_SEPARATOR = '.'.freeze
+
       # Whitelist and validate a parameter
       #
       # @param name [#to_sym] The name of the param to whitelist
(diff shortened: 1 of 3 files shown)


vyper:
Why we only use `string` for get a key?

I :heart: this approach...

```ruby
params.get('customer_name')  # => "Luca"
params.get(:customer_name)  # => "Luca"
```

joneslee85:
@vyper hash is nice but it does not represent the nested attributes very well. For that very reason, I think we should stick to string.

joneslee85:
@jodosha :+1:

PR #91, opened 2015-03-04
Added default cookies options support

AlfonsoUceda:
This allow to send default cookies values to every cookie.

Values set for a cookie by user are respected.

### DIFF
--- a/lib/lotus/action/cookie_jar.rb
+++ b/lib/lotus/action/cookie_jar.rb
@@ -36,9 +36,10 @@ module Lotus
       # @return [CookieJar]
       #
       # @since 0.1.0
-      def initialize(env, headers)
-        @_headers = headers
-        @cookies  = Utils::Hash.new(extract(env)).symbolize!
+      def initialize(env, headers, default_cookies_options)
+        @_headers                = headers
+        @cookies                 = Utils::Hash.new(extract(env)).symbolize!
+        @default_cookies_options = default_cookies_options
       end
 
       # Finalize itself, by setting the proper headers to add and remove
@@ -50,7 +51,7 @@ module Lotus
       #
       # @see Lotus::Action::Cookies#finish
       def finish
-        @cookies.each {|k,v| v.nil? ? delete_cookie(k) : set_cookie(k, v) }
+        @cookies.each { |k,v| v.nil? ? delete_cookie(k) : set_cookie(k, _merge_default_values(v)) }
       end
 
       # Returns the object associated with the given key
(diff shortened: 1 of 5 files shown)


lib/lotus/action/cookie_jar.rb
```diff
@@ -77,6 +78,21 @@ def []=(key, value)
...
 
       private
+
+      # Merge default cookies options with values provided by user
+      #
+      # Cookies values provided by user are respected
+      #
+      # @since x.x.x
+      # @api private
+      def _merge_default_values(value)
```
AlfonsoUceda: I don't like a lot this method. any idea?

lib/lotus/controller/configuration.rb
```diff
@@ -558,14 +592,15 @@ def duplicate
...
-        @default_headers    = {}
-        @action_module      = ::Lotus::Action
+        @handle_exceptions       = true
+        @handled_exceptions      = {}
+        @modules                 = []
+        @formats                 = DEFAULT_FORMATS.dup
+        @default_format          = nil
+        @default_charset         = nil
+        @default_headers         = {}
+        @default_cookies_options = {}
```
jodosha: This should be `{httponly: true}`, so we enforce a secure policy by default.

lib/lotus/action/cookie_jar.rb
jodosha: What's the thing that you don't like?

lib/lotus/controller/configuration.rb
AlfonsoUceda: do you think this should be here instead lotus gem?

lib/lotus/action/cookie_jar.rb
AlfonsoUceda: the use of is_a? method, but I think I dont have more options

lib/lotus/controller/configuration.rb
jodosha: Oh, you're right. Sorry about the confusion. This is good :+1:

lib/lotus/controller/configuration.rb
```diff
@@ -498,6 +498,39 @@ def default_headers(headers = nil)
...
+      #       domain: 'lotusrb.org',
+      #       path: '/controller',
+      #       secure: true,
+      #       httponly: true
+      #     })
+      #   end
+      def default_cookies_options(options = nil)
+        if options
+          @default_cookies_options.merge!(
+            options.reject {|_,v| v.nil? }
```
joneslee85: nitpick, please add space after `{`

joneslee85:
:+1:

PR #94, opened 2015-03-10
Fix redirect_to inside callback and doesn't call `call` method in action

### LINKED ISSUE #93: #call method is called even when a before callback does a redirect_to.
Hi everyone!

I'm experimenting some random behaviors.

I have a `before :authenticate!` callback that redirect to '/login' when a session doesn't exist. Simple.

When making the tests for the action I wanted to check that it returns 302 when user isn't  authenticated.

When running the tests, the tests were running the #call method and calling the Repository when it should stop the chain and redirect.

Here some examples.

```ruby
module Web::Controllers::Posts
  class Index
    include Web::Action
    # include Web::AuthenticatedAction

    before :authenticate!

    def call(params)
      puts  "This should not be called"
    end

    def authenticate!
      puts "Redirect should stop here"
      redirect_to '/login'
    end
  end
end
````

Here is the output when I request the controller route

```sh
[2015-03-09 18:04:15] INFO  WEBrick 1.3.1
[2015-03-09 18:04:15] INFO  ruby 2.1.4 (2014-10-27) [x86_64-darwin14.0]
[2015-03-09 18:04:15] INFO  WEBrick::HTTPServer#start: pid=49653 port=2300
Redirect should stop here
This should not be called
::1 - - [09/Mar/2015:18:04:18 -0400] "GET / HTTP/1.1" 302 - 0.9164
::1 - - [09/Mar/2015:18:04:19 -0400] "GET /login HTTP/1.1" 200 - 0.8267
```

The same apply for tests.

AlfonsoUceda:
fix #93 

This PR fixes a problem if a person does a `redirect_to` inside a before callback the `call` method in an action is called.

### DIFF
--- a/test/integration/full_stack_test.rb
+++ b/test/integration/full_stack_test.rb
@@ -81,4 +81,11 @@ describe 'Full stack application' do
       'book.author.name' => [Lotus::Validations::Error.new('book.author.name', :presence, true, nil)]
     })
   end
+
+  it "redirect in before action and call action method doesn't called" do
+    get 'users/1'
+
+    last_response.status.must_equal 302
+    last_response.body.must_equal 'Found' # This message is 302 status
+  end
 end
--- a/lib/lotus/action/redirect.rb
+++ b/lib/lotus/action/redirect.rb
@@ -12,12 +12,14 @@ module Lotus
 
       private
 
-      # Redirect to the given URL
+      # Redirect to the given URL and halt the request
       #
       # @param url [String] the destination URL
       # @param status [Fixnum] the http code
       #
       # @since 0.1.0
+      #
+      # @see Lotus::Action::Throwable#halt
       #
       # @example With default status code (302)
       #   require 'lotus/controller'
@@ -50,7 +52,7 @@ module Lotus
       #   action.call({}) # => [301, {'Location' => '/articles/23'}, '']
       def redirect_to(url, status: 302)
         headers[LOCATION] = url
-        self.status = status
+        halt(status)
       end
     end
   end
--- a/lib/lotus/action/session.rb
+++ b/lib/lotus/action/session.rb
@@ -112,8 +112,8 @@ module Lotus
       #   # The validation errors caused by Comments::Create are available
       #   # **after the redirect** in the context of Comments::Index.
       def redirect_to(*args)
-        super
         flash[ERRORS_KEY] = errors.to_a unless params.valid?
+        super
       end
 
       # Read errors from flash or delegate to the superclass
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -1050,6 +1050,29 @@ module FullStack
         end
       end
     end
+
+    module Users
+      class Show
+        include FullStack::Action
+
+        before :redirect_to_root
+        after :set_body
+
+        def call(params)
+          self.body = "call method shouldn't be called"
+        end
+
+        private
+
+        def redirect_to_root
+          redirect_to '/'
+        end
+
+        def set_body
+          self.body = "after callbacl shouldn't be called"
+        end
+      end
+    end
   end
 
   class Renderer
@@ -1079,6 +1102,10 @@ module FullStack
           post '/1', to: 'poll#step1'
           get  '/2', to: 'poll#step2'
           post '/2', to: 'poll#step2'
+        end
+
+        namespace 'users' do
+          get '/1', to: 'users#show'
         end
       end
 


lib/lotus/action/callbacks.rb
```diff
@@ -192,7 +192,7 @@ module InstanceMethods
         # @api private
         def call(params)
           _run_before_callbacks(params)
-          super
+          super if location_header.nil? || location_header.empty?
```
joneslee85: Seems to be good candidate for blank? method for utils

joneslee85:
Looks good to me

lib/lotus/action/callbacks.rb
AlfonsoUceda: I agree, I could add to https://github.com/lotus/utils/blob/master/lib/lotus/utils/string.rb

AlfonsoUceda:
@donperi mmmm maybe we need, what do you need guys?

joneslee85:
@AlfonsoUceda I have no strong opinion on that, I leave it to @jodosha

gotjosh:
I think it should not be called. If you have an expensive operation that should be called *after*, it makes no sense to execute that if you redirected. 

To me, you shouldn't get to that stage where you halted the execution chain yet you had a callback fired for no reason.

jodosha:
@AlfonsoUceda Thanks for taking care of this :smile_cat: 

My opinion on this is to keep callbacks as simple as possible.
What if we do:

```ruby
def redirect_to(url, status: 302)
  headers[LOCATION] = url
  halt(status)
end
```

This will prevent `#call` to be invoked, but it keeps the callbacks as they are.

AlfonsoUceda:
ok!, I'm goint to try that ;)

AlfonsoUceda:
@jodosha there is a problem doing that.

This test [1] fails, errors aren't passed 

[1] https://github.com/lotus/controller/blob/master/test/integration/full_stack_test.rb#L26 with

1) Failure:
Full stack application#test_0003_in case of redirect and invalid params, it passes errors in session and then deletes them [/home/alfonso/Proyectos/lotus/forks/controller/test/integration/full_stack_test.rb:31]:
Expected "FullStack::Controllers::Books::Index {:params=>#<FullStack::Controllers::Books::Index::Params:0x00000002f021f0 @env={\"rack.version\"=>[1, 3], \"rack.input\"=>#<StringIO:0x00000002f03668>, \"rack.errors\"=>#<StringIO:0x00000002f03708>, \"rack.multithread\"=>true, \"rack.multiprocess\"=>true, \"rack.run_once\"=>false, \"REQUEST_METHOD\"=>\"GET\", \"SERVER_NAME\"=>\"example.org\", \"SERVER_PORT\"=>\"80\", \"QUERY_STRING\"=>\"\", \"PATH_INFO\"=>\"\", \"rack.url_scheme\"=>\"http\", \"HTTPS\"=>\"off\", \"SCRIPT_NAME\"=>\"/books\", \"CONTENT_LENGTH\"=>\"0\", \"rack.test\"=>true, \"REMOTE_ADDR\"=>\"127.0.0.1\", \"HTTP_REFERER\"=>\"http://example.org/books\", \"HTTP_HOST\"=>\"example.org\", \"HTTP_COOKIE\"=>\"rack.session=BAh7BkkiD3Nlc3Npb25faWQGOgZFVEkiRWM4Y2Y1MWU0N2M5YTM1ZmJkNjcx%0AZDQ4YzJjMDllMGY3OTEzMDEzZGVmZDYyNjc4ZTFkNjZmYzAwZmVkZjg1NjgG%0AOwBG%0A--255d74ddc88955ac6299e6cf8b2934e95cde2410\", \"rack.session\"=>{\"session_id\"=>\"c8cf51e47c9a35fbd671d48c2c09e0f7913013defd62678e1d66fc00fedf8568\"}, \"rack.session.options\"=>{:path=>\"/\", :domain=>nil, :expire_after=>nil, :secure=>false, :httponly=>true, :defer=>false, :renew=>false, :sidbits=>128, :secure_random=>SecureRandom, :secret=>\"4efacb79d69e9eb1d6594acaa618033b\", :coder=>#<Rack::Session::Cookie::Base64::Marshal:0x00000002b355b8>}, \"router.request\"=>#<HttpRouter::Request:0x00000002f02da8 @rack_request=#<Rack::Request:0x00000002f02dd0 @env={...}>, @path=[], @extra_env={}, @params=[], @acceptable_methods=#<Set: {\"GET\", \"HEAD\"}>>, \"router.params\"=>{}, \"rack.request.query_string\"=>\"\", \"rack.request.query_hash\"=>{}, \"rack.request.cookie_hash\"=>{\"rack.session\"=>\"BAh7BkkiD3Nlc3Npb25faWQGOgZFVEkiRWM4Y2Y1MWU0N2M5YTM1ZmJkNjcx\\nZDQ4YzJjMDllMGY3OTEzMDEzZGVmZDYyNjc4ZTFkNjZmYzAwZmVkZjg1NjgG\\nOwBG\\n--255d74ddc88955ac6299e6cf8b2934e95cde2410\"}, \"rack.request.cookie_string\"=>\"rack.session=BAh7BkkiD3Nlc3Npb25faWQGOgZFVEkiRWM4Y2Y1MWU0N2M5YTM1ZmJkNjcx%0AZDQ4YzJjMDllMGY3OTEzMDEzZGVmZDYyNjc4ZTFkNjZmYzAwZmVkZjg1NjgG%0AOwBG%0A--255d74ddc88955ac6299e6cf8b2934e95cde2410\", \"rack.session.unpacked_cookie_data\"=>{\"session_id\"=>\"c8cf51e47c9a35fbd671d48c2c09e0f7913013defd62678e1d66fc00fedf8568\"}}, @raw=#<Lotus::Utils::Attributes:0x00000002f01e08 @attributes={}>, @attributes=#<Lotus::Utils::Attributes:0x00000002f01e08 @attributes={}>, @errors=#<Lotus::Validations::Errors:0x00000002f00df0 @errors={}>>, :errors=>#<Lotus::Validations::Errors:0x00000002f00df0 @errors={}>, :format=>:all}" to include "@actual=\"\"".

AlfonsoUceda:
@jodosha nevermind I got it, I'll update the PR

joneslee85:
:+1:

test/integration/full_stack_test.rb
```diff
@@ -81,4 +81,11 @@ def app
       'book.author.name' => [Lotus::Validations::Error.new('book.author.name', :presence, true, nil)]
     })
   end
+
+  it "redirect in before action and call action method doesn't called" do
+    get 'users/1'
+
+    last_response.status.must_equal 302
+    last_response.body.must_equal 'Found' # This message is 302 status
```
jodosha: @AlfonsoUceda Can you please ensure that the body isn't send, as it's useless?

```ruby
last_response.body.must_equal ''
```

AlfonsoUceda: ok @jodosha ;) I'll do another PR

gotjosh:
This looks great to me!

PR #96, opened 2015-03-12
Make flash session to persist between requests. Fix #95

### LINKED ISSUE #95: Flash session is nil when doing redirect.
Hello everyone.

I'm having an issue with flash session. I'm setting the flash session and then using a redirect_to.
In the next request I'm expecting to get the value but is returning nil.

```ruby
  def call(params)
      @organization = Organization.new(params[:organization])

      if @repository.persist(@organization)
        flash[:success] = "Organization created"
        redirect_to routes.path(:manager_organizations)
      else 
        flash[:errors] = @errors = @organization.errors
      end
  end
```

In the view:
```haml
 - if flash[:success]
   = flash[:success]
```

The code is working when the flash is used in the same request. (Using the flash[:errors] in this case.)

donperi:
This fix issue #95

### DIFF
--- a/lib/lotus/action/flash.rb
+++ b/lib/lotus/action/flash.rb
@@ -12,6 +12,13 @@ module Lotus
       # @api private
       SESSION_KEY = :__flash
 
+      # Session key where the last request_id is stored
+      #
+      # @since x.x.x
+      # @api private
+
+      LAST_REQUEST_KEY = :__last_request_id
+
       # Initialize a new Flash instance
       #
       # @param session [Rack::Session::Abstract::SessionHash] the session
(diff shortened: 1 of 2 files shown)


lib/lotus/action/flash.rb
```diff
@@ -136,6 +145,29 @@ def remove!
...
         flash.values
       end
+
+
+      # Get the last request session flash
+      #
+      # @return [Hash] the flash of last request
+      # @since x.x.x
+      # @api private
+
```
joneslee85: nitpick, please remove empty line

lib/lotus/action/flash.rb
```diff
@@ -136,6 +145,29 @@ def remove!
...
+        flash[@last_request_id] || {}
+      end
+
+      # Store the last request_id to create the next flash with its values
+      # is current flash is not empty.
+
+      # @return [void]
+      # @since x.x.x
+      # @api private
+
```
joneslee85: nitpick, please remove empty line

lib/lotus/action/flash.rb
```diff
@@ -12,6 +12,13 @@ class Flash
       # @api private
       SESSION_KEY = :__flash
 
+      # Session key where the last request_id is stored
+      #
+      # @since x.x.x
+      # @api private
+
```
joneslee85: nitpick, please remove empty line

joneslee85:
looks legit to me

joneslee85:
@donperi let's wait for feedback from @jodosha

jodosha:
@donperi Thanks for this PR, it looks good. :+1: 
I'd still keep flash private API for now. It still needs to get more and more mature. /cc @joneslee85

PR #97, opened 2015-03-15
Force empty message when use redirect_to

AlfonsoUceda:
complete #94


jodosha:
@AlfonsoUceda Thank you for this extra PR :+1: 

May I ask you to verify two things with `Rack::Lint`?

  1. Does it complains if we send an empty string?
  2. What happens if we set the body as `nil`?

Thank you :smiley_cat:

jodosha:
@AlfonsoUceda I was wrong. 302 can have a body, there is no mention to suppress it. So there is no need to have this PR.

Sorry for the confusion. :cry:

AlfonsoUceda:
ok thanks @jodosha  ;)

PR #98, opened 2015-03-15
Allow blocks when specifying Rack middleware to use

Erol:
I am using Lotus::Controller outside of a Lotus application and ran into a problem when trying to use Rack::Auth::Basic. I was trying to configure it just like below:

```
  prepare do
    use Rack::Auth::Basic, "Protected Area" do |username, password|
      username == 'admin' && password == 'password'
    end
  end
end
```

But I noticed that `Lotus::Action::Rack::ClassMethods#use` does not accept blocks when specifying middleware. I checked https://github.com/rack/rack/blob/master/lib/rack/builder.rb#L81 and `Rack::Builder#use` does accept blocks.

Is this intended behavior? Am I missing something?

### DIFF
--- a/lib/lotus/action/rack.rb
+++ b/lib/lotus/action/rack.rb
@@ -94,8 +94,8 @@ module Lotus
         #       end
         #     end
         #   end
-        def use(middleware, *args)
-          rack_builder.use middleware, *args
+        def use(middleware, *args, &block)
+          rack_builder.use middleware, *args, &block
         end
       end
 
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -170,6 +170,18 @@ class ExposeAction
   end
 end
 
+class ZMiddleware
+  def initialize(app, &message)
+    @app = app
+    @message = message
+  end
+
+  def call(env)
+    code, headers, body = @app.call(env)
+    [code, headers.merge!('Z-Middleware' => @message.call), body]
+  end
+end
+
 class YMiddleware
   def initialize(app)
     @app = app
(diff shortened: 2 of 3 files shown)


Erol:
Also, can't find the test for `Lotus::Action::Rack::ClassMethods#use`. Should I make one? :hamster:

jodosha:
@Erol You're right, it should accept a block.

### Action.use

`Action.use` is intended for fine grained middleware configuration. For instance, if only a few resources need to use a Rack component, but not the entire app. When used, the action has a small perf penalty, but this is a good compromise because incoming requests won't go through the middleware when not needed.

Eg.

```ruby
class RuntimeAction
  include Lotus::Action
  use Rack::Runtime

  def call(params)
    # ...
  end
end

class DashboardAction
  include Lotus::Action

  def call(params)
    # ...
  end
end

# GET /runtime # small perf penalty because of .use
# GET /dashboard # perf is OK, because it doesn't go through Rack::Runtime
```

### Configuration#prepare

On the other hand, `Configuration#prepare` is intended for general configurations (not fine grained). Configuring `.use` for all the action, will add that small perf penalty to all the actions.

### Solution

If you need to filter **all** the requests for **all** the actions (no fine grained approach), I suggest to mount that middleware in `config.ru`

jodosha:
@Erol Please add tests here `test/integration/use_test.rb`, and I'll be happy to merge this. Thank you :+1:

Erol:
@jodosha Done. And thanks for the `config.ru` tip!

jodosha:
@Erol Thanks for the quick implementation. Merged :+1:

PR #105, opened 2015-05-04
Fix internet explorer HTTP accept

### LINKED ISSUE #104: Wrong format on Internet Explorer
Howdy fellow lotus lovers!

I noticed that two of my Lotus Production apps won't allow Internet Explorer access. They would immediately throw a 500 error when trying to access them. After some digging, these are my findings:

The `HTTP_ACCEPT` header that IE (short for Internet Explorer from now on) is coming with; is `text/html, application/xhtml+xml, image/jxr, */*` when checking `self.format` inside an controller action we can see that this translates to `:'123'`. Weird right?

After digging around in `lotus/controller` I noticed this particular method inside [mime.rb](https://github.com/lotus/controller/blob/master/lib/lotus/action/mime.rb#L418)

````ruby
def accepts
  unless accept == DEFAULT_ACCEPT
    ::Rack::Utils.best_q_match(accept, ::Rack::Mime::MIME_TYPES.values)
  end
end
````

Since `accept` is equal to `HTTP_HEADER` the result from Rack's `best_q_match` is `"application/vnd.lotus-1-2-3"` (more info on this [here](https://www.iana.org/assignments/media-types/application/vnd.lotus-1-2-3)) which is what is ultimatetly determining the not-so-cool `:'123'` format.

# SIDENOTE
I took a few side steps onto Rack to check out this [function](https://github.com/rack/rack/blob/master/lib/rack/utils.rb#L200-L224) and this is the result from `q_values` for this particular string:

`[["text/html", 1.0], ["application/xhtml+xml", 1.0], ["image/jxr", 1.0], ["*/*", 1.0]]`

Seems weird that the quality of all of them is 1.0 considering that `text/html` should be the one to be respected, in particular `*/*` is the one being extrapolated to `"application/vnd.lotus-1-2-3"`, I know that our very own @jodosha patched this function and added a test case that involves this particular string so I believe something broke in Rack along the way. You can take a look at @jodosha's patch [here](https://github.com/rack/rack/commit/e06a25423979e0443e6252cbe0cb4b2b41fa89f0)

I want to help out here, so my question is, is this something that should be patched/fixed in Lotus or completely delegate this to Rack? It's believe is quite bad for us (Lotus) that IE is broken out of the box.

jodosha:
Closes #104


gotjosh:
Great decision to remove `rack-patch.rb` IMHO and exceptionally documenting this problem. Great work as always @jodosha :tophat:

jodosha:
@gotjosh Thank you, Sir :tophat:

PR #108, opened 2015-05-14
Rack::Lint compatibility

AlfonsoUceda:
## Problem

Both `Rack::Lint` and Rack SPEC aren't compliant with RFC 2016 (see https://github.com/rack/rack/issues/787) for entity headers.

The RFC allows headers like `Content-Type` and `Content-Length` to be sent when the status code is `204`, but `Rack::Lint` don't.

## Solution

We make Lotus compliant with `Rack::Lint` until they don't fix this problem.

### DIFF
--- a/lib/lotus/action/head.rb
+++ b/lib/lotus/action/head.rb
@@ -30,11 +30,9 @@ module Lotus
         'Allow'            => true,
         'Content-Encoding' => true,
         'Content-Language' => true,
-        'Content-Length'   => true,
         'Content-Location' => true,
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@weppos is this fix your problem? @jodosha told me about your problem

jodosha:
@AlfonsoUceda This fixes the problem, thank you :+1:

weppos:
Thanks for your work! :clap: It is working without issues now.

PR #110, opened 2015-05-15
Added automatically expires option if max_age option presents

### LINKED ISSUE #107: Cookies: automatic "expires" setting
Old browsers (IE 6, 7, 8), don't support `max-age` setting.

If we set default `:max_age` from `Lotus::Configuration`, or we specify inline in a cookie `cookie[:foo] = {value: 'bar', max_age: 300}`, it will be ignored.

I think it's our role to provide transparent solutions to browsers quirks.
We should set `expires` automatically, when it isn't present, but max-age is.

```ruby
# Do nothing
cookies[:foo] = 'bar'

# Do nothing
cookies[:foo] = { value: 'bar', expires: Time.parse('2015-05-14 12:01:00') }

# Automatically set expires 
cookies[:foo] = { value: 'bar', max_age: 120 }
```

See http://mrcoles.com/blog/cookies-max-age-vs-expires/

AlfonsoUceda:
Closes #107 

Now when you set max_age in a cookie, automatically expires option is set (Time.now + max_age option).
```ruby
cookies[:foo] = { value: 'bar', max_age: 120 }
```
```ruby
"'Set-Cookie' => 'bar=foo; max-age=120; expires=Fri, 15 May 2015 10:49:42 -0000'"
```

### DIFF
--- a/lib/lotus/action/cookie_jar.rb
+++ b/lib/lotus/action/cookie_jar.rb
@@ -98,11 +98,21 @@ module Lotus
       # @api private
       def _merge_default_values(value)
         cookies_options = if value.is_a? Hash
-          value
+          value.merge(_add_expires_option(value) || {})
         else
           { value: value }
         end
         @default_options.merge cookies_options
+      end
+
+      # Add expires option to cookies if :max_age presents
+      #
+      # @since x.x.x
+      # @api private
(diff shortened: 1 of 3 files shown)


jodosha:
@AlfonsoUceda thank you :clap:

PR #112, opened 2015-05-21
Added send_file method in action for sending files

### LINKED ISSUE #102: Send file
Add a `send_file` method to actions, in order to send files to HTTP clients.
Quick and dirty implementation available at: https://github.com/lotus/controller/tree/send-file

AlfonsoUceda:
closes #102 

# Problem
Lotus hasn't a send file feature

# Solution
```ruby
require 'lotus/controller'

class Show
  include Lotus::Action

  def call(params)
    # ...
    senf_file Pathname.new('path_to_file')
  end
end 
```


jodosha:
@AlfonsoUceda This is good :+1: Before to merge it, I'd love to see more test cases for the following scenarios:

  * Forced `Content-Type` (eg, client sends `Accept: text/html`, but the file is a PNG, so it must return `Content-Type: application/png`.
  * HEAD request
  * Conditional GET
  * HTTP_RANGE (See `Rack::Utils.byte_ranges`)

test/assets/test.txt
```diff
@@ -0,0 +1,2 @@
+Mock Text File
+This file is used to help with senf file test feature.
```
pel-daniel: typo in 'senf' => 'send'

AlfonsoUceda:
@jodosha can you check if these tests are correct?

test/integration/send_file_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      last_response.status.must_equal 200
+      last_response.headers['Content-Length'].to_i.must_equal file.size
+      last_response.headers['Content-Type'].must_equal 'image/png'
+    end
+  end
+
+  describe 'conditional get request' do
+    it "shouldn't send file" do
+      get '/files/1', {}, 'HTTP_ACCEPT' => 'text/html', 'HTTP_IF_MODIFIED_SINCE' => Time.now.gmtime.rfc2822
```
AlfonsoUceda: This test return 200, I think should return 304 with `HTTP_IF_MODIFIED_SINCE` header because it's a conditional request

jodosha: It returns `200` because the value is wrong. `Time.now` will be always greater than the file creation date :wink:

AlfonsoUceda: You are right, I don't think about that :(

jodosha:
@AlfonsoUceda Thank you, I've just merged it :+1:

Issue #113, opened 2015-05-27
Lotus::Action::Parameters converting to hash


bradrobertson:
right now if I have something like:

```ruby
class LotusParams < Lotus::Action::Params
  param :contact do
    param :email, presence: true, format: /.+@.+/
  end
end
```

And I call
```ruby
p = LotusParams.new(contact: {email: 'xxx'})
p.to_h
```
I get
```ruby
{"contact"=>#<#<Class:0x007fe5e222f798>:0x007fe5e219d7d0 @env={"email"=>"xxx"}, @raw=#<Lotus::Utils::Attributes:0x007fe5e219d690 @attributes={"email"=>"xxx"}>, @attributes=#<Lotus::Utils::Attributes:0x007fe5e219d4b0 @attributes={"email"=>"xxx"}>>}
```

I'd like to actually convert this to a raw hash of ruby primitives as specified in these params data types. I looked at the `raw` method, but that doesn't return the whitelist or anything, it's just the ones passed in.

Is there an easy way to convert a params object back to a validated, whitelisted raw hash?

bradrobertson:
If you're looking for context here on the usage, I want to start using Lotus in little bits as much as possible, but I'm not ready for the the framework as a whole.

My first thought is to write a little rails gem using these param validation classes that would allow some to specify the required params for a rails action and get a `validated_params` in their action method assuming it passed the conditions. These `validated_params` should ultimately be the final representation of the request params as a raw hash, rather than a `Lotus::Action::Params` object

g3d:
Got a similar error with passing params into initializer for worker/service.

mjbellantoni:
Pull #115 has a proposed fix for this.

jodosha:
Fixed by #115

PR #115, opened 2015-06-11
Fixes bug #113 where Params are incorrectly converted to hashes.

mjbellantoni:
The problem occurs when we're using whitelisting and nested
parameters.  When attribute definitions are created in the context
of Lotus::Action::Params the anonymous Params subclass which is
created does not respond to the :lotus_nested_attributes? method
which is expected by Lotus::Utils::Attribues#to_h.

The fix is to make the anonymous class respond to that method with
the right value.  This is a little hacky for two reasons:

1. You need to look at the source for
Lotus::Validations::AttributeDefiner to understand this.
2. It might make sense to create a NestParams class in
Lotus::Action.

### DIFF
--- a/lib/lotus/action/params.rb
+++ b/lib/lotus/action/params.rb
@@ -119,7 +119,11 @@ module Lotus
       # @since 0.3.2
       # @api private
       def self.build_validation_class(&block)
-        kls = Class.new(Params)
+        kls = Class.new(Params) do
+          def lotus_nested_attributes?
+            true
+          end
+        end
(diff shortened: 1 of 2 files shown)


mjbellantoni:
Aside: I noticed both `Lotus::Action::Params#{to_h,to_hash}` are aliased to each other.  Test for each appear in the test suite but they're not identical.  Is this intentional?  Should one be removed or should they be made identical?

jodosha:
@mjbellantoni Just merged this, thank you :+1:

PR #116, opened 2015-06-26
Officially support rubinius

tak1n:


AlfonsoUceda:
build passes, :+1: to me

joneslee85:
This to me is a big decision, firstly we do want to support Rbx but we are afraid that changes in rbx might break our build again and having more platform to support is not a viable option for our scarce resources atm. If anyone from rbx could be committed into continuing this support, I'd vote :+1: to me

tak1n:
I am more or less "anyone from rbx" (https://github.com/orgs/rubinius/people?utf8=%E2%9C%93&query=tak1n) and keen on continuing the effort to support lotus on rbx.

Also `lotus-validations` already officially supports rbx https://github.com/lotus/validations#rubies
Feel free to close, having said that I will continue to get lotus working on rbx :smiley:

AlfonsoUceda:
@tak1n thank you ;)

jodosha:
@tak1n Thank you for this, but I'm still not keen to support Rubinius officially. We're still on a phase where we want to move fast and MRI is the main reference for us.
When we were trying to catch with it in the past, even Rubinius patch releases suddenly broke our builds.

Before 1.0 we'll decide for real which implementations to support. For Rubinius it will depend on its stability.

jodosha:
@tak1n @lotus/core-team I'm merging this with the purpose of observing Rubinius stability. If in the future it breaks the build please move it back to allowed failures.

tak1n:
@jodosha I understand, thx for merging.
Personally I will continue to do work on this and we can see before 1.0 if it is worth for lotus to support rbx or not.

Cheers :smile:

PR #119, opened 2015-08-17
handles inherited exceptions with specified method

manuwell:
Hi guys, 

This PR ables lotus to handle inherited exceptions. This is important if we want to handle unknow exceptions. 

What do you think?

Thks


rpanachi:
Excellent @manuwell, this is exactly what I needed!

PR #125, opened 2015-09-15
introduces request and response format and deprecates default_format

### LINKED ISSUE #111: should default_format :json also force self.format = :json?
Currently the response format does not follow `default_format`. 
Maybe it should use `default_format` for response also and the user should have to override if they want something other then the default?

theocodes:
Recap..

As per my issue #111  and suggestion from Luca, this PR introduces `default_request_format` and `response_request_format` and deprecates `default_format`. 

The premise behind it is:

`default_request_format` behaves the same as `default_format`. However, now it's clear what default format we're setting. request..

`default_response_format` will set it's value to all responses. It can only be overwritten by directly setting `self.format = :whatever` in the action itself. If a value is not set, it will then behave as it did before (try to guess response format based on request format etc...)

`default_format` has been deprecated and should be removed eventually.

Please have a look and let me know what you think. :-)

Notes:

* There's a PR on lotus/lotus that addresses the same thing.
* I've versioned the new methods to 0.5.0 as I assume that's when it will be introduced, but that needs to be confirmed. thoughts?

Closes #111

### DIFF
--- a/lib/lotus/controller/configuration.rb
+++ b/lib/lotus/controller/configuration.rb
@@ -418,35 +418,92 @@ module Lotus
       # an argument, it will set the corresponding instance variable. When
       # called without, it will return the already set value, or the default.
       #
-      # @overload default_format(format)
+      # @overload default_request_format(format)
       #   Sets the given value
       #   @param format [#to_sym] the symbol format
       #   @raise [TypeError] if it cannot be coerced to a symbol
       #
-      # @overload default_format
+      # @overload default_request_format
       #   Gets the value
       #   @return [Symbol,nil]
       #
-      # @since 0.2.0
       #
       # @see Lotus::Action::Mime
       #
       # @example Getting the value
       #   require 'lotus/controller'
       #
-      #   Lotus::Controller.configuration.default_format # => nil
+      #   Lotus::Controller.configuration.default_request_format # => nil
       #
       # @example Setting the value
       #   require 'lotus/controller'
       #
       #   Lotus::Controller.configure do
-      #     default_format :html
+      #     default_request_format :html
       #   end
-      def default_format(format = nil)
+      #
+      # @since 0.5.0 TODO: Confirm?
+      def default_request_format(format = nil)
         if format
-          @default_format = Utils::Kernel.Symbol(format)
+          @default_request_format = Utils::Kernel.Symbol(format)
         else
-          @default_format
+          @default_request_format
+        end
+      end
+
+      # Set a format as default fallback for all the requests without a strict
+      # requirement for the mime type.
+      #
+      # @since 0.2.0
+      #
+      # @deprecated Use {#default_request_format} instead.
+      def default_format(format = nil)
+        Lotus::Utils::Deprecation.new('default_format is deprecated, please use default_request_format')
+        default_request_format(format)
+      end
+
+      # Set a format to be used for all responses regardless of the request type.
+      #
+      # The given format must be coercible to a symbol, and be a valid mime type
+      # alias. If it isn't, at the runtime the framework will raise a
+      # `Lotus::Controller::UnknownFormatError`.
+      #
+      # By default this value is nil.
+      #
+      # This is part of a DSL, for this reason when this method is called with
+      # an argument, it will set the corresponding instance variable. When
+      # called without, it will return the already set value, or the default.
+      #
+      # @overload default_response_format(format)
+      #   Sets the given value
+      #   @param format [#to_sym] the symbol format
+      #   @raise [TypeError] if it cannot be coerced to a symbol
+      #
+      # @overload default_response_format
+      #   Gets the value
+      #   @return [Symbol,nil]
+      #
+      #
+      # @see Lotus::Action::Mime
+      #
+      # @example Getting the value
+      #   require 'lotus/controller'
+      #
+      #   Lotus::Controller.configuration.default_response_format # => nil
+      #
+      # @example Setting the value
+      #   require 'lotus/controller'
+      #
+      #   Lotus::Controller.configure do
+      #     default_response_format :json
+      #   end
+      #
+      # @since 0.5.0 TODO: Confirm?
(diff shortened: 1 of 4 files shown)


test/configuration_test.rb
```diff
@@ -324,21 +346,21 @@ def hash
...
-      @config.send(:formats).must_equal     @configuration.send(:formats)
-      @config.default_format.must_equal     @configuration.default_format
-      @config.default_charset.must_equal    @configuration.default_charset
-      @config.default_headers.must_equal    @configuration.default_headers
+      @config.handle_exceptions.must_equal      @configuration.handle_exceptions
+      @config.handled_exceptions.must_equal     @configuration.handled_exceptions
+      @config.action_module.must_equal          @configuration.action_module
+      @config.modules.must_equal                @configuration.modules
+      @config.send(:formats).must_equal         @configuration.send(:formats)
+      @config.default_request_format.must_equal @configuration.default_request_format
```
jodosha: Can you please test `default_response_format` too? :smile:

test/configuration_test.rb
```diff
@@ -347,7 +369,7 @@ def hash
       @config.action_module    CustomAction
       @config.prepare          { include Comparable }
       @config.format another: 'another/format'
-      @config.default_format :json
+      @config.default_request_format :json
```
jodosha: Same as: https://github.com/lotus/controller/pull/125/files#r39599214

test/configuration_test.rb
```diff
@@ -356,7 +378,7 @@ def hash
       @config.action_module.must_equal               CustomAction
       @config.modules.size.must_equal                2
       @config.format_for('another/format').must_equal :another
-      @config.default_format.must_equal               :json
+      @config.default_request_format.must_equal               :json
```
jodosha: Same as: https://github.com/lotus/controller/pull/125/files#r39599214

test/configuration_test.rb
```diff
@@ -365,7 +387,7 @@ def hash
       @configuration.action_module.must_equal      ::Lotus::Action
       @configuration.modules.size.must_equal       1
       @configuration.format_for('another/format').must_be_nil
-      @configuration.default_format.must_equal     :html
+      @configuration.default_request_format.must_equal :html
```
jodosha: Same as: https://github.com/lotus/controller/pull/125/files#r39599214

test/configuration_test.rb
```diff
@@ -378,7 +400,7 @@ def hash
       @configuration.action_module    CustomAction
       @configuration.modules          { include Kernel }
       @configuration.format another: 'another/format'
-      @configuration.default_format :another
+      @configuration.default_request_format :another
```
jodosha: Same as: https://github.com/lotus/controller/pull/125/files#r39599214

test/configuration_test.rb
```diff
@@ -391,7 +413,7 @@ def hash
       @configuration.action_module.must_equal(::Lotus::Action)
       @configuration.modules.must_equal([])
       @configuration.send(:formats).must_equal(Lotus::Controller::Configuration::DEFAULT_FORMATS)
-      @configuration.default_format.must_be_nil
+      @configuration.default_request_format.must_be_nil
```
jodosha: Same as: https://github.com/lotus/controller/pull/125/files#r39599214

test/integration/mime_type_test.rb
```diff
@@ -111,6 +123,12 @@ def call(params)
     response.body.must_equal                    'html'
   end
 
+  it 'uses default_response_format if set in the configuration regardless of request format' do
+    response = @app.get('/response')
+    response.headers['Content-Type'].must_equal 'application/json; charset=utf-8'
+    response.body.must_equal                    'html'
+  end
+
```
jodosha: I think we should also assert that `self.format=` will be honored.

```ruby
class OverrideDefaultResponse
  include Lotus::Action

  configuration.default_response_format :json

  def call(params)
    self.body = :xml
  end
end

it 'allows to override default_response_format' do
  response = @app.get('/overwritten_format')
  response.headers['Content-Type'].must_equal 'application/xml; charset=utf-8'
end
```

jodosha:
> I've versioned the new methods to 0.5.0 as I assume that's when it will be introduced, but that needs to be confirmed. thoughts?

This is fine right now, but please use `@since x.x.x` instead. Before a release we grep for these YARD tags and replace with the right value :smile:

jodosha:
@theocodes This looks good, I left a few comments, can you please address these issues?

test/configuration_test.rb
```diff
@@ -187,25 +187,47 @@ def hash
     end
   end
 
-  describe '#default_format' do
```
jodosha: @theocodes We need both the tests for `#default_format` and `#default_request_format`. The first is still there, and we want to make sure that it still works. :wink:

lib/lotus/controller/configuration.rb
```diff
@@ -418,35 +418,92 @@ def format(hash)
...
+      end
+
+      # Set a format as default fallback for all the requests without a strict
+      # requirement for the mime type.
+      #
+      # @since 0.2.0
+      #
+      # @deprecated Use {#default_request_format} instead.
       def default_format(format = nil)
+        Lotus::Utils::Deprecation.new('default_format is deprecated, please use default_request_format')
```
jodosha: @theocodes We probably need a test case for this deprecation. You can use Minitest `capture_io`.

```ruby
_, err = capture_io do
    default_format :png
  end
end

err.chomp.must_equal "..."
```

test/integration/mime_type_test.rb
theocodes: Agreed!  Thanks  :+1:

theocodes:
@jodosha I believe I have addressed all the issues.  :)

Note: There's a lot of noise about that deprecated method when you run the test suite now.. Is there way to mute it for tests only?

jodosha:
@theocodes This looks good, thank you :+1:

PR #126, opened 2015-09-16
Bump rack dep to ~> 1.6 + >= 1.6.2 because of CVE-2015-3225

jodosha:
See:
https://groups.google.com/forum/#!topic/rubyonrails-security/gcUbICUmKMc


joneslee85:
:+1:

PR #128, opened 2015-10-19
Make session public

cfcosta:
This enables setting and testing session attributes.


jodosha:
@cfcosta Merged, thank you! :+1:

PR #129, opened 2015-10-27
Added Streaming Response example

berfarah:
Hey @jodosha, added an example that I would use for a streaming response. I think a SSE example would be good too, but I don't have much experience with those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -921,6 +921,37 @@ action.call({ 'HTTP_ACCEPT' => '*/*' }) # => Content-Type 'application/custom'
 action.format                           # => :custom
 ```
 
+### Streamed Responses
+
+When the work to be done by the server takes time, it may be a good idea to stream your response. Here's an example of a streamed CSV.
+
+```ruby
+Lotus::Controller.configure do
+  format csv: 'text/csv'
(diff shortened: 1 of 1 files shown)


jodosha:
@berfarah Thank you, this looks good. Would you mind to describe which web servers work with it. From the description we communicate that WEBrick doesn't work, so we should provide a reliable alternative (aka standard) for this. What do you use?

berfarah:
@jodosha I've added puma to the readme, though I believe unicorn, thin and passenger would work as well. Would you like to provide multiple options or just the one?

joneslee85:
@berfarah I think we should stick to `puma`, as it is recommended on Heroku and elsewhere too

berfarah:
@joneslee85 Sounds good. In the README I have now:
[Streaming does not work with WEBrick as it buffers its response. We recommend using `puma`, though you may find success with other servers ](https://github.com/lotus/controller/pull/129/files#diff-04c6e90faac2675aa89e2176d2eec7d8R953)

joneslee85:
good to me

jodosha:
@berfarah Thank you! :+1:

PR #131, opened 2015-11-18
adding extra conditions where actions need instantiation

baob:
Updating the readme to be consistent with the advice provided in https://github.com/lotus/controller/issues/130


jodosha:
@baob Thanks for this addition to README :+1:

PR #132, opened 2015-11-27
Allow to force custom headers for responses that shouldn't include them

jodosha:
According to RFC 2616, when a response MUST have an empty body, it only allows Entity Headers.

For instance, a <tt>204</tt> doesn't allow <tt>Content-Type</tt> or any other custom header.

This restriction is enforced by <tt>Lotus::Action::Head#finish</tt>.

However, there are cases that demand to bypass this rule to set meta
informations via headers.

An example is a <tt>DELETE</tt> request for a JSON API application. It returns a <tt>204</tt> but still wants to specify the rate limit quota via <tt>X-Rate-Limit</tt>.

```ruby
require 'lotus/controller'

module Books
  class Destroy
    include Lotus::Action

    def call(params)
      # ...
      self.headers.merge!(
        'Last-Modified' => 'Fri, 27 Nov 2015 13:32:36 GMT',
        'X-Rate-Limit'  => '4000',
        'Content-Type'  => 'application/json',
        'X-No-Pass'     => 'true'
      )

      self.status = 204
    end

    private

    def keep_response_header?(header)
      super || header == 'X-Rate-Limit'
    end
  end
end
```

Only the following headers will be sent:

  * `Last-Modified` - because we used `super' in the method that respects the HTTP RFC
  * `X-Rate-Limit`  - because we explicitely allow it

Both `Content-Type` and `X-No-Pass` are removed because they're not allowed

If there are no objections, I would release `v0.4.6` with only this change, because needed by a project I'm working on.

/cc @lotus/core-team

### DIFF
--- a/lib/lotus/action/head.rb
+++ b/lib/lotus/action/head.rb
@@ -50,7 +50,7 @@ module Lotus
 
         if _requires_no_body?
           @_body = nil
-          @headers.reject! { |header,_| !ENTITY_HEADERS.include?(header) }
+          @headers.reject! {|header,_| !keep_response_header?(header) }
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/lotus/action/head.rb
```diff
@@ -50,7 +50,7 @@ def finish
 
         if _requires_no_body?
           @_body = nil
-          @headers.reject! { |header,_| !ENTITY_HEADERS.include?(header) }
+          @headers.reject! {|header,_| !keep_response_header?(header) }
```
AlfonsoUceda: space between `{` and `|`

AlfonsoUceda:
:+1:

joneslee85:
@jodosha :+1:

PR #139, opened 2016-01-11
Exception handling: ensure superclass exception to not shadow subclasses

jodosha:
## The Problem

This fixes the following use case.

We have a custom error that inheriths from `StandardError`

```ruby
class MyCustomError < StandardError
end
```

In the `prepare` block we add a handler for generic errors `StandardError`

```ruby
  prepare do
    handle_exception StandardError => 500
  end
end
```

In an action we have an handler for our custom error. 

```ruby
class MyAction
  include Lotus::Action
  handle_exception MyCustomError => 400

  def call(params)
    # ...
  end
end
```

Because we added first the handler for `StandardError`, we have the following handled exceptions for `MyAction`:

```ruby
{StandardError=>500, MyCustomError=>400}
```

Now, because `StandardError` is a superclass for `MyCustomError`, it shadows our error because of this change introduced by #119 

```
# Lotus::Controller::Configuration

      def exception_handler(exception)
        handler = nil

        @handled_exceptions.each do |exception_class, h|
          if exception.kind_of?(exception_class)
            handler = h
            break
          end
        end

        handler || DEFAULT_ERROR_CODE
      end
```

When `MyCustomError` is raised, `exception.is_a?(exception_class)` returns `true`, because of the exception hierarchy described above.

## The Fix

We sort internally the exceptions according to their hierarchy. That means the handled exceptions for `MyAction` will now be:

```ruby
{MyCustomError=>400, StandardError=>500}
```

When we look for the proper handler, we'll find first the subclass.

---
Ref: #119
/cc @lotus/core-team @manuwell @weppos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Complete, fast and testable actions for Rack
 
 ### Fixed
 - [Cainã Costa] Ensure Rack environment to be always available for sessions unit tests
+- [Luca Guidi] Ensure superclass exceptions to not shadow subclasses during exception handling (eg. `CustomError` handler will take precedence over `StandardError`)
 
 ### Changed
 - [Luca Guidi] Removed `Lotus::Controller::Configuration#default_format`
--- a/lib/lotus/controller/configuration.rb
+++ b/lib/lotus/controller/configuration.rb
@@ -180,6 +180,7 @@ module Lotus
       #   end
       def handle_exception(exception)
         @handled_exceptions.merge!(exception)
+        _sort_handled_exceptions!
       end
 
       # Return a callable handler for the given exception
@@ -682,6 +683,13 @@ module Lotus
       end
 
       protected
+      # @since 0.5.0
+      # @api private
+      def _sort_handled_exceptions!
+        @handled_exceptions = Hash[
+          @handled_exceptions.sort{|(ex1,_),(ex2,_)| ex1.ancestors.include?(ex2) ? -1 : 1 }
+        ]
+      end
 
       attr_accessor :handled_exceptions
       attr_accessor :formats
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -126,8 +126,8 @@ end
 class ErrorCallFromInheritedErrorClassStack
   include Lotus::Action
 
-  handle_exception MyCustomError => :handler
   handle_exception StandardError => :standard_handler
+  handle_exception MyCustomError => :handler
 
   def call(params)
     raise MyCustomError


AlfonsoUceda:
:+1:

davydovanton:
cool! :+1:

weppos:
@jodosha I see your approach is to sort all the exceptions each time a new handler is added. Wouldn't it be better to simply maintain the exceptions in a sorted Array and add a new exception in the proper position?

That would avoid you to sort all N exceptions each time you add a new one, relying on the fact that the array is already sorted.

jodosha:
@weppos Hey thanks for this suggestion :+1:

We're gonna skip this improvement for now because we wanted it to work first, before tomorrow's release (v0.5.0). The perf gain isn't too big because we're speaking about a few entries in that Hash, and it's a win that would only affect startup time.

However, I'm pinging @davydovanton if he wants to speedup it. Target should be the next release (0.5.1?).

Thank you all! :+1:

davydovanton:
yes, I can try to work on this improvement :smiley:

weppos:
Great. @davydovanton feel free to ping me on the new PR if you want an extra review.

davydovanton:
@weppos I implement two different ways to shorting hash. Benchmark you can find [here](https://gist.github.com/davydovanton/bacf1b675ddd259b73ca). What do you think? The `third_method` is what do you represent?

weppos:
Correct. I also wonder if you have to keep storing the values in a `Hash`. It seems to me that if you use an Array of Array representation

```ruby
[
  [CustomError, :whatever_custom],
  [StandardError, :whatever_standard],
]
```

you may benefit from converting back and forth from Hash to Array. What is the benefit of storing the list of exception is a Hash?

The only possible reason is that you need a fast lookup in case of exception. Is it the case?

weppos:
According [to this method](https://github.com/lotus/controller/blob/7fa013b6bcf2cdbd590fd4dfb3b79285d44e1877/lib/lotus/controller/configuration.rb#L215-L218), this seems to be the case.

However, I also wonder: if you perform an exact lookup using a hash, what will happen if:

- I add a handler for StandardError
- My controller raises FooError that inherit from StandardError

Will the error be rescued or not? In other word, do you perform exact-class rescue, or also inheritance rescue?

It seems to me the current code performs the former, whereas Ruby the latter

```ruby
class FooError < StandardError
end

begin
    raise FooError
rescue StandardError => e
    # will rescue
end
```

davydovanton:
@weppos yes, your code is correct. But what I need to do if I need execute different code for each error?
``` ruby
class FooError < StandardError
end

class BarError < StandardError
end

begin
  rand(20 - 10) > 0 ? raise FooError : raise BarError
rescue StandardError => e
  # wrong
  # I need special code for FooError and for BarError
end
```

weppos:
What you generally do in Ruby is

```ruby
begin
  rand(20 - 10) > 0 ? raise FooError : raise BarError
rescue FooError
  # 
rescue BarError
  # 
end
```

or 

```ruby
begin
  rand(20 - 10) > 0 ? raise FooError : raise BarError
rescue FooError
  # 
rescue # fallback anything else to StandardError
  # 
end
```

davydovanton:
yes, I know about this :)
So, what about improving performance? I think that current solution is better. What do you think @weppos?

weppos:
> So, what about improving performance? I think that current solution is better. What do you think @weppos?

`third_method` seems accurate.

PR #140, opened 2016-01-13
Fixed `Set-Cookie` header send twice in second and following request

### LINKED ISSUE #138: with session + cookie enabled, set-cookie header for rack.session is sent twice
When having ```cookies``` and ```sessions``` enabled, then the __set-cookie__ header for __rack.session__ is sent twice, once with and once without __path__.
This leads to confusing state when modifying session content.

Steps to reproduce:

1) create a new lotus app
2) modify files as follows
```ruby
# application.rb: enable cookies AND sessions
...
cookies true
sessions :cookie, secret: 'verrrrrry-secret'
..

# routes.rb
get '/', to: 'home#index'

# web/controllers/home/index.rb
module Web::Controllers::Home

  class Index
    include Web::Action

    def call(params)
      self.body = 'Welcome'
    end
  end
end
```
3) lotus server
4) visit localhost:2300 and check response headers, you'll see

```
Set-Cookie:rack.session=BLA...BLA; HttpOnly
Set-Cookie:rack.session=BLA...BLA; path=/; HttpOnly
```

First cookie is from ```cookies``` second one from ```sessions```

Problem can be "fixed" by setting the __path__ of ```cookies``` to the same value as the __path__ of ```sessions``` but still two headers with same content are sent back to the browser.

AlfonsoUceda:
This PR resolves #138 

### The problem
If we have cookies and sessions enabled in `application.rb` like:
```ruby
cookies true
sessions :cookie, secret: 'SECRET'
```
and we make a first request to one endpoint of our Application we have the following cookie in the response:
```ruby
Set-Cookie: rack.session=hash; path="/"; HttpOnly;
```
and the cookie session is set in out browser.

If we make a second request, then the following header in request is sent:
```ruby
Cookie: rack.session=hash
```
and the following headers in the response:
```ruby
Set-Cookie: rack.session=hash; path="/"; HttpOnly;
Set-Cookie: rack.session=hash; path="/"; HttpOnly;
```
Because the Cookie is sent in the second request and followings we shouldn't get `Set-Cookie` headers for `rack.session`.

---

cc @lotus/core-team

### DIFF
--- a/lib/lotus/action/cookie_jar.rb
+++ b/lib/lotus/action/cookie_jar.rb
@@ -16,6 +16,12 @@ module Lotus
       # @api private
       HTTP_HEADER       = 'HTTP_COOKIE'.freeze
 
+      # The key used by Rack to set the session cookie
+      #
+      # @since x.x.x
+      # @api private
+      RACK_SESSION_KEY   = :'rack.session'
+
       # The key used by Rack to set the cookies as an Hash in the env
       #
       # @since 0.1.0
@@ -55,6 +61,7 @@ module Lotus
       #
       # @see Lotus::Action::Cookies#finish
       def finish
+        @cookies.delete(RACK_SESSION_KEY)
         @cookies.each { |k,v| v.nil? ? delete_cookie(k) : set_cookie(k, _merge_default_values(v)) }
       end
 
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -1300,3 +1300,110 @@ class HandledRackExceptionAction
     raise TestException.new
   end
 end
+
+module SessionWithCookies
+  Controller = Lotus::Controller.duplicate(self) do
+    handle_exceptions false
+
+    prepare do
+      include Lotus::Action::Glue
+      include Lotus::Action::Session
+      include Lotus::Action::Cookies
+    end
+  end
+
+  module Controllers
+    module Home
+      class Index
+        include SessionWithCookies::Action
+
+        def call(params)
+        end
+      end
+    end
+  end
+
+  class Renderer
+    def render(env, response)
+      action = env.delete('lotus.action')
+
+      if response[0] == 200 && action.renderable?
+        response[2] = "#{ action.class.name } #{ action.exposures }"
+      end
+
+      response
+    end
+  end
+
+  class Application
+    def initialize
+      resolver = Lotus::Routing::EndpointResolver.new(namespace: SessionWithCookies::Controllers)
+      routes   = Lotus::Router.new(resolver: resolver) do
+        get '/',     to: 'home#index'
+      end
+
+      @renderer   = Renderer.new
+      @middleware = Rack::Builder.new do
+        use Rack::Session::Cookie, secret: SecureRandom.hex(16)
+        run routes
+      end
+    end
+
+    def call(env)
+      @renderer.render(env, @middleware.call(env))
+    end
+  end
+end
+
+module SessionsWithoutCookies
+  Controller = Lotus::Controller.duplicate(self) do
+    handle_exceptions false
+
+    prepare do
+      include Lotus::Action::Glue
+      include Lotus::Action::Session
+    end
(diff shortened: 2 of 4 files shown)


lib/lotus/action/cookie_jar.rb
```diff
@@ -131,6 +137,7 @@ def extract(env)
         hash   = env[COOKIE_HASH_KEY] ||= {}
         string = env[HTTP_HEADER]
 
+        return if string && string.include?(RACK_SESSION_KEY)
```
pascalbetz: @AlfonsoUceda this will delete any cookie that contains RACK_SESSION_KEY in either _key_ or _value_

```ruby
# in your action
...
cookies['message'] = "This will be deleted because it contains 'rack.session'"
...
```

test/fixtures.rb
```diff
@@ -1300,3 +1300,110 @@ def call(params)
     raise TestException.new
   end
 end
+
+module CookiesAndSessions
```
pascalbetz: Nitpicking: i'd call this module and corresponding test: ```SessionWithCookies```
Like that it is in sync with ```SessionWithoutCookies```

test/integration/cookies_and_sessions_test.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def app
+    CookiesAndSessions::Application.new
+  end
+
+  def response
+    last_response
+  end
+
+  it 'passes action inside the Rack env' do
```
pascalbetz: i don't understand the name of the test.

AlfonsoUceda: sorry, copy/paste, I have to modify this >_<

test/fixtures.rb
AlfonsoUceda: :+1:

lib/lotus/action/cookie_jar.rb
AlfonsoUceda: yes, you are right. the `Set-Cookie` header value for rack session is `rack.session=hash`. maybe string isn't the correct var to check, maybe it's hash, what do you think @pascalbetz ?

pascalbetz: I'm not entirely sure:-(

Perhaps we can skip setting a cookie with key  ```rack.session``` in ```CookieJar``` altogether? Or only when ```sessions``` is enabled with ```:cookie```?

AlfonsoUceda: there isn't communication between sessions middleware and cookies jar (now), I'll try to inspect in the bottom of this method how the cookie is converted with `Rack::Utils`

lib/lotus/action/cookie_jar.rb
```diff
@@ -131,6 +137,7 @@ def extract(env)
         hash   = env[COOKIE_HASH_KEY] ||= {}
         string = env[HTTP_HEADER]
 
+        return if hash.has_key? RACK_SESSION_KEY
```
jodosha: @AlfonsoUceda Isn't this `nil` return value dangerous for the caller? It expects a Hash, but get `nil` if already the session was already set.

jodosha: @AlfonsoUceda does it means that if `rack.session` key, the cookie jar will appear as empty? Can we improve the tests and let the actions to set both cookies and sessions values? Thank you :+1:

jodosha: @AlfonsoUceda I tried it manually with a test application. This fix introduces another problem: it works fine for the first request, but for the second one `Set-Cookie` header is **NOT** sent at all.

/cc @pascalbetz

AlfonsoUceda: In a previous version I checked string instead hash with has_key, I'll try to update test with string version and adding more cookies in tests

jodosha:
@AlfonsoUceda This works great, thank you :+1:

PR #145, opened 2016-03-22
Fix Ruby warnings

jodosha:
/cc @hanami/core-team @hanami/contributors

### DIFF
--- a/hanami-controller.gemspec
+++ b/hanami-controller.gemspec
@@ -26,5 +26,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler',   '~> 1.6'
   spec.add_development_dependency 'minitest',  '~> 5'
   spec.add_development_dependency 'rack-test', '~> 0.6'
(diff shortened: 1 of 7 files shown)


davydovanton:
why did you set  `@cache_control_directives` and `@expires_directives` variables to `nil` ?

jodosha:
@davydovanton Due the dynamic nature of Ruby, a non-initialized variable is equivalent to `nil`. But referencing non-initialized instance variables emits a warning.

```ruby
class Foo
  def bar
    @bar || 42
  end
end
```

To fix the warning:

```ruby
class Foo
  def initialize
    @bar = nil
  end

  def bar
    @bar || 42
  end
end
```

The public API to set that value
is defined by https://github.com/hanami/controller/blob/fix-ruby-warnings/lib/hanami/action/cache/cache_control.rb#L28 and https://github.com/hanami/controller/blob/fix-ruby-warnings/lib/hanami/action/cache/cache_control.rb#L32

PR #148, opened 2016-04-01
FIX: set only changed cookies

ADone:
`Hanami::Action::CookieJar` sets up response header `Set-Cookie` for **all** incoming cookies at current path/domain.
And CookieJar apply same rules (`httponly` or `secure`) to all setted cookies, that unfortunately breaks some counters and third-party metrics.
I tried to avoid this side effect by setting only changed cookies in response.


jodosha:
@ADone Thanks for this fix, I just merged it. :+1:

PR #149, opened 2016-04-05
Makes errors in session.rb public in order to be testable 

### LINKED ISSUE #147: NoMethodError: private method `errors' called for action
Reported by @nickmorlan in chat.

The problem is caused by `session.rb`, that defines `errors` as a private method.

Here's a stack trace:
https://gist.github.com/nickmorlan/4ce60b1632f115152289

vasspilka:
Fixes #147

### DIFF
--- a/lib/hanami/action/session.rb
+++ b/lib/hanami/action/session.rb
@@ -61,6 +61,19 @@ module Hanami
         @_env[SESSION_KEY] ||= {}
       end
 
+      # Read errors from flash or delegate to the superclass
+      #
+      # @return [Hanami::Validations::Errors] A collection of validation errors
+      #
+      # @since 0.3.0
+      # @api private
+      #
+      # @see Hanami::Action::Validatable
+      # @see Hanami::Action::Session#flash
+      def errors
+        flash[ERRORS_KEY] || super
+      end
+
       private
 
       # Container useful to transport data with the HTTP session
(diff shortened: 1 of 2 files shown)


jodosha:
@vasspilka Hello and thanks for this fix. Before to merge this PR, we need to add a test. The test will prove that the patch solves the problem **and** it will prevent this issue to come back again (regression).

Would you consider to add enhance this test case? https://github.com/hanami/controller/blob/master/test/method_visibility_test.rb

Thanks.

vasspilka:
Will this do?

cllns:
Hey @vasspilka, I think the test might be better in `test/session_test.rb`. What do you think?

Also, rather than testing to see if the public method exists, what about testing the return value of that method?
Similar to what is done in [`test/action_test.rb`](https://github.com/hanami/controller/blob/7e45605903beba8a30e9499cf1b8ceea4f74c88f/test/action_test.rb#L33-L39) (you might have to add some params validation to `SessionAction` in `test/fixtures.rb` in order to get an error?)

vasspilka:
It returns an empty array when everything is ok, should I create a new `@action` and then check if its `#error` are `#empty?` ?

cllns:
@vasspilka I think adding an actual error to the session is best, to ensure it's getting passed through. Otherwise, the `errors` method might be coming from just `Action`, not `Action::Sessions`

jodosha:
@vasspilka I just merged this, thanks!

PR #160, opened 2016-06-29
Remove global configuration

jodosha:
## Adios, Global Configuration

This is a huge, but sane change for Hanami: remove global configurations.

`Hanami::Controller` module will not hold a global configuration anymore. Single applications will hold their own separated, immutable configuration.

``` ruby
require 'hanami/controller'

module Web
  module Controllers
    include Hanami::Controller

    configure do |config|
      # ...
    end
  end
end

module Admin
  module Controllers
    include Hanami::Controller

    configure do |config|
      # ...
    end
  end
end
```

**NOTE:** `Web::Controllers` and `Admin::Controllers` modules will hold their own separated configuration. All the actions declared them will inherit a copy of the configuration.
### No More Namespace Depth Restrictions

This could be an unknown fact, but we had a strict requirement for Ruby namespace structure with a depth of 4 constants: `Web::Controllers::Articles::Show`. With this change we can leave freedom to structure the code as developers want.

For instance:

``` ruby
module API
  module V2
    module Controllers
      include Hanami::Controller
    end
  end
end
```

This will result in `API::V2::Controllers::Articles::Show`. **Please note that it’s not our intention to change how Hanami full stack apps work, this is only for developers who want to use `hanami-controller` in standalone mode.**
#### For Us

While developing this framework we always suffered from problems due to the global configuration. We wanted to test how an action inherits the exception handling, so we had to change settings into `Hanami::Controller.configuration` and remember to immediately restore the previous value… and yet we had intermittent failures for long time.

In testing code now we use scoped configurations to simulate real world applications (eg. `Web::Controllers` holds its own configuration).
#### For Developers

With this proposal there is an hidden, but useful feature: configuration is **injected** in action instances.

If a developer wants to unit test an action under different conditions, now they can specify the settings in the constructor.

``` ruby
action = Web::Controllers::Articles::Show.new(configuration: { handle_exceptions: false })
```

Before they needed to prepare and restore the global configuration in setup/teardown testing code. Which use to make brittle their builds.
## No More “Framework Duplication”

"Framework duplication" is removed. It was a technique to duplicate modules like `Hanami::Action` into `Web::Action` in order to allow multiple instances of the framework to coexist within the same Ruby process.

This technique was:
- Dangerous because it propagates global configurations to duped modules
- Source of confusion for new developers: “Where `Web::Action` is defined?”

With this proposed change all the actions will always include `Hanami::Action`.

That means for Hanami full stack projects no more `Web::Action` or `Admin::Action`: actions will always include `Hanami::Action`.
## Immutable Configurations

Immutability is an insurance that prevents accidental changes at the runtime.

Technically speaking it saves us a lot of memory because we don’t need to copy global configuration to all the actions.
Only the actions that specify different settings will get a copy of the original configuration.

So if your application has 100 actions and just 5 have different settings, the framework will reuse the same configuration instance for all that 95 remaining actions.
### Syntax Changes

Immutability made necessary some syntax changes to keep configuration implementation simple. Now there isn’t a way to apply changes via `.configuration`, but via a `configure` block.

This makes impossible to change the configuration from the outside of the action, but it’s still possible to read it.

This has the benefit of moving away from DSLish syntax where the same method is used both to read and write values.

``` ruby
module Web::Controllers::Articles
  class Show
    include Hanami::Action
    configuration.handle_exceptions false
  end
end
```

We want to change in favor of real setters and getters:

``` ruby
module Web::Controllers::Articles
  class Show
    include Hanami::Action

    configure do |config|
      config.handle_exceptions = false
    end
  end
end
```

---

If we think that this is a good idea, we will apply the same changes to all the other Hanami frameworks.
The final goal is to make more solid the Hanami boot process.

---

/cc @hanami/core-team @hanami/contributors @solnic @AMHOL @timriley


solnic:
This is **a very reasonable change**. I know lots of rubyists are used to global configs but it's a huge anti-pattern that unfortunately was widely adopted. FWIW I tried to use some of hanami components last year and the global config turned out to be a blocker, as I wanted to have 2 configurations of the same component and it just didn't work. So yeah, this is great :)

Reducing global state is *in general* a good direction, regardless of the library/framework. Once you can easily instantiate components of your system and pass things around without being worried about silly side-effects (ie some config being changed during the lifecycle of your system, either at boot time or, even worse, run-time!) your life becomes simpler; and as you pointed out - this doesn't just make developing libs/frameworks simpler, but also has benefits for *the users*.

👏 👍

AlfonsoUceda:
@jodosha will this PR be available before 1.0.0? could you rebase stable and resolve conflicts when you can? thanks ;)

jodosha:
@AlfonsoUceda This will be part of 2.0. I'm closing this for now.

PR #164, opened 2016-07-27
Add documentation for params#valid?

davydovanton:
/cc @joneslee85

### DIFF
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -90,6 +90,15 @@ module Hanami
         end
       end
 
+      # Returns true if no validation errors are found,
+      # false otherwise.
+      #
+      # @return [TrueClass, FalseClass]
+      #
+      # @since 0.7.0
+      #
+      # @example
+      #   params.valid? # => true
       def valid?
         @result.success?
       end


lib/hanami/action/params.rb
```diff
@@ -90,6 +90,14 @@ def error_messages(error_set = errors)
         end
       end
 
+      # Returns true if params is valid
+      #
+      # @return [Boolean]
```
joneslee85: s/ with [TrueClass, FalseClass]

because there is no Boolean type in Ruby

lib/hanami/action/params.rb
```diff
@@ -90,6 +90,14 @@ def error_messages(error_set = errors)
         end
       end
 
+      # Returns true if params is valid
```
joneslee85: What if params in not invalid? What do you think about restructuring this doc comment to reveal the intention of the method rather than the specific implementation behaviour?

joneslee85:
ping @davydovanton

davydovanton:
@joneslee85 updated

davydovanton:
thanks

PR #165, opened 2016-07-27
Refactor error_messages method

davydovanton:
/cc @joneslee85

### DIFF
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -77,16 +77,14 @@ module Hanami
       #   params.error_messages
       #     # => ["Email is missing", "Email is in invalid format", "Name is missing", "Tos is missing", "Age is missing", "Address is missing"]
       def error_messages(error_set = errors)
-        error_set.each_with_object([]) do |(key, messages), result|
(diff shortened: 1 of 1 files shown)


joneslee85:
@davydovanton thanks for the PR, would you be able to provide a micro benchmark to compare the perf? That looks good to me but I think it would be more convincing for the community if the PR has clear benchmarking results.

jodosha:
@davydovanton @joneslee85 I ran the benchmarks and the refactoring doesn't improve the speed:

```
  Warming up --------------------------------------
      each_with_object   465.000  i/100ms
              flat_map   450.000  i/100ms
  Calculating -------------------------------------
      each_with_object      4.598k (± 6.2%) i/s -     23.250k in   5.079603s
              flat_map      4.602k (± 4.1%) i/s -     23.400k in   5.093111s

  Comparison:
              flat_map:     4602.1 i/s
      each_with_object:     4597.7 i/s - same-ish: difference falls within error
```

See the code at: https://gist.github.com/jodosha/d457b44ee60744b754a843b7d3f8d54f

PR #168, opened 2016-08-01
Allow uploaded Files in whitelisted params

### LINKED ISSUE #166: Tempfile possibly being stringified during param whitelisting
Hallo,

Congratulations on the new release!

While upgrading to 0.8 a file upload feature started to break. On looking more closely, it seemed that when a multi part file upload is made from a form and then checked with a dry-v validation, the value of of a file parameter is stringified.

https://github.com/hanami/controller/blob/master/lib/hanami/action/params.rb#L129

This is the example we have been trying out:

https://github.com/hanami/controller#validations--coercions

I created a little app to see what kind of types were being delivered in the params with and without validations. It seems without validations the tempfile is not being modified.

https://github.com/matthewling/test-file-upload-hanami

Thanks for your help!

Matthew

fidothe:
The implementation of `params` is stringifying all values prior to
sending them to validation, which means that if you use params
whitelisting with multipart form uploads your Tempfile objects get
turned into strings of the form `"<#File:0x123456>"`, effectively
breaking form uploads for anyone who's restricting their params.

This PR just adds another branch to case statement in `stringify!`
to allow `File`. This is obviously a fairly brittle approach, but after
some discussion in the dry-v gitter channel it seems like this may be a
useful temporary fix until a better approach can be found.

closes #166


fidothe:
cc @solnic

joneslee85:
Thanks @fidothe for the PR

I haven't been involved with the dry-v change in params code so please bear with me while I am reviewing the PR. The one thing that get my attention is the test, to me it is very confusing, I would like to see the test cover all the coercion logic in params. And lastly I am questing if we should keep the `::File` type params as is, maybe the params should be the full filepath to the file instead? That would keep the value pretty consistent with the method intention, that is to `stringify` the params. Thoughts?

fidothe:
The problem with stringifying a `File` object is that you then lose access to the File object, effectively making it impossible to have a form that has both whitelisted params and an uploaded file.

This PR is not a good fix for the problem – there's clearly some deeper issue with the way that dry-v and hanami form validation are being used together (see the chat in dry-rb's gitter room here https://gitter.im/dry-rb/chat?at=579f1e9c9684847b374240d6), but it would allow us to deploy Hanami 0.8 in production, which we really want to, but can't while all our file uploading is now broken. (The text in #166 is a little diplomatic – this is a showstopper breaking change for us.)

The long term solution is to address the concerns of that `FIXME` comment here https://github.com/hanami/controller/blob/master/lib/hanami/action/params.rb#L112, and I'm trying to get up to speed with dry-v and the new validations logic so I can help with that, but I really want to be able to deploy the JSON logging stuff without abandoning all form param checking in the meantime...

cllns:
@fidothe Would #172 solve this?

joneslee85:
@fidothe I put this PR on hold until Luca is back from holiday, we don't know yet why he introduced this concept.

fidothe:
@cllns yes, but i think it would break other stuff – the params are definitely being stringified for a reason…

joneslee85:
@fidothe https://github.com/hanami/controller/pull/172 has removed the params stringfication, can you please update the PR accordingly?

jodosha:
@fidothe Thanks for this PR. With #172 we removed the useless params stringification. I merged this PR because it adds new tests to prevent regressions in the future. Thank you!

/cc @joneslee85 @cllns

PR #170, opened 2016-08-02
Use default content type when accepting `*/*`

### LINKED ISSUE #167: Format for IE8
Hi 
I posted this in the hanami chat posting here for collecting all information;

I was wondering if someone can give me a little pointer here as I’m not familiar with how the self.format is being selected.

My app works fine in most browsers and gets the format :html matching text/html from the HTTP_ACCEPT except in IE8.

It keeps requesting :jpeg with the error

`Hanami::View:MissingTemplateError: Can’t find template ’signup/new’ for ‘jpeg’ format.`

Since IE8 has `*/*` at the end of its HTTP_ACCEPT should it not also get :html?

I saw the issue [#104](https://github.com/hanami/controller/issues/104) raised for wrong IE format and the fix [#105](https://github.com/hanami/controller/pull/105)
This was added to version 0.4.0, I have version 0.6.1 running.

I tried to debug `best_q_match` in  `lib/lotus/action/mime.rb`it seems like `*/*` is being resolved as `application/vnd.lotus-1-2-3`, since it is the first in the list of available mime types.

amccann1:
`*/*` is being resolved as `application/vnd.lotus-1-2-3`, since it is
the first in the list of available mime types.

This sets it to the default content type in these situations

fixes #167

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -487,9 +487,13 @@ module Hanami
       # @see https://github.com/hanami/controller/issues/104
       def best_q_match(q_value_header, available_mimes)
         values = ::Rack::Utils.q_values(q_value_header)
-
         values = values.map do |req_mime, quality|
-          match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
+          if req_mime == DEFAULT_ACCEPT
+             # See https://github.com/hanami/controller/issues/167
+             match = default_content_type
+          else
+             match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
+          end
           next unless match
           [match, quality]
         end.compact
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -489,10 +489,10 @@ module Hanami
         values = ::Rack::Utils.q_values(q_value_header)
         values = values.map do |req_mime, quality|
           if req_mime == DEFAULT_ACCEPT
-             # See https://github.com/hanami/controller/issues/167
-             match = default_content_type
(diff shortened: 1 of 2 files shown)


lib/hanami/action/mime.rb
```diff
@@ -487,9 +487,13 @@ def content_type_with_charset
...
       def best_q_match(q_value_header, available_mimes)
         values = ::Rack::Utils.q_values(q_value_header)
-
         values = values.map do |req_mime, quality|
-          match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
+          if req_mime == DEFAULT_ACCEPT
+             # See https://github.com/hanami/controller/issues/167
+             match = default_content_type
+          else
+             match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
```
hieuk09: I think the code here has 3-spaces tab. Could you re-indent it?

lib/hanami/action/mime.rb
```diff
@@ -487,9 +487,13 @@ def content_type_with_charset
       # @see https://github.com/hanami/controller/issues/104
       def best_q_match(q_value_header, available_mimes)
         values = ::Rack::Utils.q_values(q_value_header)
-
         values = values.map do |req_mime, quality|
-          match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
+          if req_mime == DEFAULT_ACCEPT
+             # See https://github.com/hanami/controller/issues/167
```
hieuk09: @joneslee85: should we put the issue link here?

jodosha:
@amccann1 Thanks for this fix! 👍

PR #172, opened 2016-08-02
Do not stringify params

Moratorius:
The params do not need to be stringified for dry-v (and can cause issues with parameter types that should not be coerced).
See https://gitter.im/dry-rb/chat?at=579f20f5f1da4f376e1e9da9.

### DIFF
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -106,31 +106,8 @@ module Hanami
 
       private
 
-      # @since 0.7.0
-      # @api private
-      def _extract_params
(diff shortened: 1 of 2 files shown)


joneslee85:
@Moratorius I would put this on hold until @jodosha come back to clarify the intent on stringfying params.

jodosha:
@Moratorius @joneslee85 @solnic I introduced that private method to make the test to pass with `dry-validation` while it was on high development (pre 0.8, master). I'm really glad we can remove that code. Thanks all! 👏

PR #173, opened 2016-08-02
Do not symbolize raw request params

Moratorius:
The raw request params should not be changed.

### DIFF
--- a/lib/hanami/action/base_params.rb
+++ b/lib/hanami/action/base_params.rb
@@ -45,7 +45,7 @@ module Hanami
       def initialize(env)
         @env    = env
         @raw    = _extract_params
-        @params = Utils::Hash.new(@raw).symbolize!.to_h
+        @params = Utils::Hash.new(@raw).deep_dup.symbolize!.to_h
         freeze
       end
 
--- a/test/action/base_params_test.rb
+++ b/test/action/base_params_test.rb
@@ -5,6 +5,14 @@ describe Hanami::Action::BaseParams do
     @action = Test::Index.new
   end
 
+  describe '#initialize' do
+    it 'creates params without changing the raw request params' do
+      env = { 'router.params' => { 'some' => { 'hash' => 'value' } } }
+      @action.call(env)
+      env['router.params'].must_equal({ 'some' => { 'hash' => 'value' } })
+    end
+  end
+
   describe '#valid?' do
     it 'always returns true' do
       @action.call({})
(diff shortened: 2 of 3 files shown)


lib/hanami/action/base_params.rb
```diff
@@ -45,7 +45,7 @@ class BaseParams
       def initialize(env)
         @env    = env
         @raw    = _extract_params
-        @params = Utils::Hash.new(@raw).symbolize!.to_h
+        @params = Utils::Hash.new(@raw).deep_dup.symbolize!.to_h
```
duykhoa: I think there is a way to improve this code, since it looks quite complicated and not good in term of performance 😺

jodosha:
@Moratorius Hey thanks for this PR. It totally makes sense.

I tried to merge it locally and to solve the git conflicts, but there are now tests that fail. Would you mind to take care of them? Thanks!

Moratorius:
@jodosha I've rebased this branch onto master, but the failing test (https://github.com/hanami/controller/pull/173/files#diff-08a2561671240a10f5942576de7645c6R31) brings up the question of whether or not the parameters should be `deep_dup`.  The duplicated file upload object is identical in content, but is a different object.  I'm assuming that a deep symbolize is desired here so that nested parameter access is consistent (e.g. `params[:foo][:bar]` instead of `params[:foo]['bar']` in the case of a shallow symbolize).  Perhaps the symbolization that happens should produce a copy of the raw params where only the keys are symbolized, so the non-hash values are the same (same `object_id`)?

jodosha:
@Moratorius 
> brings up the question of whether or not the parameters should be `deep_dup`

I think they should. We want to ensure a fresh copy of the objects to be `params`, so if they get manipulated (uploaded file included), `params.raw` will always keep the original reference.

> The duplicated file upload object is identical in content, but is a different object.

Yes, that is the purpose of `deep_dup`: having different memory references (aka `object_id`) for the same content.

What is tricky in that test is the file comparison. As you say "different reference, same content". Ruby makes easy this kind of comparison for other objects. See this example:

```ruby
➜ pry
[1] pry(main)> a = "foo"
"foo"
[2] pry(main)> b = a.dup
"foo"
[3] pry(main)> a.object_id
70262476171440
[4] pry(main)> b.object_id
70262476007520
[5] pry(main)> a == b
true
```

A simple `#==` is enough for `String`, while `File` needs `FileUtils.cmp`.

My point is: we should keep `deep_dup`. What do you think?

Moratorius:
I agree with using `deep_dup`, I'm just concerned that it may surprise users who aren't aware that some objects aren't deeply duplicated (such as `File` objects).

Using this test as an example, reading `@action.params[:upload]` will advance the file pointer for `@action.params.raw['upload']`.

jodosha:
@Moratorius I just merged this, thanks!

PR #174, opened 2016-08-03
Raise error on reserved word exposure.

### LINKED ISSUE #158: Raise exception for reserved words (namely `flash`)
This issue started as a request for documentation that `flash` is a reserved word when using `Hanami::Action::Session`: https://github.com/hanami/hanami/issues/595

Instead of just documenting it, we should help the developer by raising an exception when they try to use any reserved word.

One reserved word is `flash` but we should make sure all reserved words are covered.

(Thanks for the original issue @rafaels88)

akhramov:
Closes #158
(following text is extracted from [comment](https://github.com/hanami/controller/issues/158#issuecomment-234843765))

I think we can go with guard module. The idea behind is dead simple:
* Create a module implementing `#expose` method, which raises an exception if given list of words contains one or more reserved word. Otherwise it calls super.
* Prepend this module to an action.
* Introduce an alias method for `#expose` (`#expose!` or `#_expose`) for internal usage. This alias won't be watched by the guard so no exceptions are raised.

### DIFF
--- a/lib/hanami/action/exposable/guard.rb
+++ b/lib/hanami/action/exposable/guard.rb
@@ -1,0 +1,99 @@
+module Hanami
+  module Action
+    module Exposable
+      # Guard for Exposures API.
+      # Prevents exposure of reserved words
+      #
+      # @since x.x.x
+      #
+      # @see Hanami::Action::Exposable::Guard::ClassMethods#expose
+      # @see Hanami::Action::Exposable::Guard::ClassMethods#reserved_word?
+      module Guard
+        # Override Ruby's hook for modules.
+        # It prepends a guard for the exposures logic
+        #
+        # @param base [Class] the target action
+        #
+        # @since x.x.x
+        # @api private
+        #
+        # @see http://www.ruby-doc.org/core-2.1.2/Module.html#method-i-included
+        def self.included(base)
+          class << base
+            prepend ClassMethods
+          end
+        end
+
+        # Exposure of reserved words
+        #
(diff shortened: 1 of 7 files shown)


lib/hanami/action/exposable/guard.rb
```diff
@@ -0,0 +1,99 @@
...
+          # @raise [IllegalExposeError] if names contain one or more of reserved
+          #   words
+          #
+          # @since x.x.x
+          # @api private
+          def detect_reserved_words!(names)
+            names.each do |name|
+              if reserved_word?(name)
+                error_text = "#{name} is a reserved word. It cannot be exposured"
+                raise IllegalExposeError, error_text
```
cllns: We don't need to define `error_text` here. You can break it across two lines after the comma to keep it under 80 characters
```
  raise IllegalExposeError, 
    "..."
```

lib/hanami/action/exposable/guard.rb
```diff
@@ -0,0 +1,99 @@
...
+          #
+          # @raise [IllegalExposeError] if names contain one or more of reserved
+          #   words
+          #
+          # @since x.x.x
+          # @api private
+          def detect_reserved_words!(names)
+            names.each do |name|
+              if reserved_word?(name)
+                error_text = "#{name} is a reserved word. It cannot be exposured"
```
cllns: `exposured` => `exposed`

jodosha:
@akhramov Thank you for this enhancement. I'm merging it!

PR #177, opened 2016-08-30
allow action params to read into arrays

TiteiKo:
Hey all,

Hash#dig method allows the following:

``` ruby
h = { foo: { array: [{ bar: 'foobar' }] }
h.dig(:foo, :array, 0, :bar) == 'foobar'
```

This PR aims at making BaseParams#get iso with Hash#dig:

``` ruby
h = Params.new(foo: { array: [ { bar: 'foobar' } ] })
h.get('foo.array.0.bar') == 'foobar'
```

This is required for the following PR: https://github.com/hanami/helpers/pull/74

### DIFF
--- a/lib/hanami/action/base_params.rb
+++ b/lib/hanami/action/base_params.rb
@@ -93,11 +93,11 @@ module Hanami
         key, *keys = key.to_s.split(GET_SEPARATOR)
         return if key.nil?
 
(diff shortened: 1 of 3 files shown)


jodosha:
@TiteiKo This looks good. Thank you 👍

PR #179, opened 2016-09-30
Resolve Content-Type from custom and common MIME types via Accept header

russCloak:
The `#content_type` method was not taking into account when we have a custom MIME type registered for a potential controller format - instead it was overlooking these values and simply skipping to the default content type.

Now, we honor the ordered preference of choosing which content type to use:
1. Explicit set value (from #format=)
2. Weighted value from Accept header (common MIME types _and_ registered custom MIME types)
3. Configured default content type
4. Hard-coded default content type

Examples:

Given the following controller:

``` ruby
module Web
  class Application < Hanami::Application
    configure do
      controller.format v1: 'application/vnd.custom.v1+json'
      controller.format v2: 'application/vnd.custom.v2+json'
      controller.prepare do
        accept :v1, :v2, :json
      end
    end
  end
end
```

_Before this PR_
When a request was made with `Accept: application/vnd.custom.v2+json` it would respond with `Content-Type: application/json`.

_With this PR_
When a request is made with `Accept: application/vnd.custom.v2+json`, it responds with `Content-Type: application/vnd.custom.v2+json`. Likewise, `v1` returns `v1`, and `application/json` returns `application/json`.

Weighting also works now for custom types:

_Before this PR_
When a request was made with `Accept: application/json; q=0.5, application/vnd.custom.v1+json; q=0.9`, it would respond with `Content-Type: application/json`.

_With this PR_
When a request is made with `Accept: application/json; q=0.5, application/vnd.custom.v1+json; q=0.9`, it responds with `Content-Type: application/vnd.custom.v1+json`. Likewise, a request made with `Accept: application/json; q=0.8, application/vnd.custom.v1+json; q=0.4` responds with `Content-Type: application/json`.

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -215,7 +215,7 @@ module Hanami
         return @content_type if @content_type
 
         if accept_header?
-          type = get_mime_type_from_accept_header
+          type = mime_type_from_accept_header
           return type if type
         end
 
@@ -471,7 +471,7 @@ module Hanami
       # @see Hanami::Controller::Configuration#format
       #
       # @api private
-      def get_mime_type_from_accept_header
+      def mime_type_from_accept_header
         all_types = (MIME_TYPES + configuration.format_mime_types).uniq
         best_q_match(accept, all_types)
       end
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -435,10 +435,19 @@ module Hanami
         @accept ||= @_env[HTTP_ACCEPT] || DEFAULT_ACCEPT
       end
 
-      # @since 0.1.0
+      # Look at the Accept header for the current request and see if it
+      # matches any of the common MIME Types (see Hanami::Action::Mime#MIME_TYPES).
+      #
+      # @return [Nil] when the Accept header does not match any known (common)
+      # MIME Types.
+      # @return [String] the charset of the request.
+      #
+      # @since 0.7.0
+      #
+      # @see Hanami::Action::Mime#MIME_TYPES
+      #
       # @api private
-      def accepts
-        unless accept == DEFAULT_ACCEPT
+      def get_common_mime_type_from_accept_header
           best_q_match(accept, MIME_TYPES)
         end
       end
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -450,6 +450,25 @@ module Hanami
       def get_common_mime_type_from_accept_header
           best_q_match(accept, MIME_TYPES)
         end
+
+      # Look at the Accept header for the current request and see if it
+      # matches any of the custom registered MIME Types
+      # (see Hanami::Controller::Configuration#format).
+      #
+      # @return [Nil] when there is not an Accept header, or the Accept header
+      # does not match any known (common) MIME Type.
+      # @return [String] the charset of the request.
+      #
+      # @since 0.7.0
+      #
+      # @see Hanami::Controller::Configuration#format
+      #
+      # @api private
+      def get_custom_mime_type_from_accept_header
+        # If the Accept header doesn't match any custom MIME types, return nil.
+        return nil unless custom_type = configuration.format_for(accept)
+        # Return the MIME type for the custom content "type" from the configuration.
+        configuration.mime_type_for(custom_type)
       end
 
       # @since 0.5.0
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -435,6 +435,16 @@ module Hanami
         @accept ||= @_env[HTTP_ACCEPT] || DEFAULT_ACCEPT
       end
 
+      # Checks if there is an Accept header for the current request.
+      #
+      # @return [Boolean] true if there's an Accept header to look at
+      #
+      # @since 0.7.0
+      # @api private
+      def accept_header?
+        accept != DEFAULT_ACCEPT
+      end
+
       # Look at the Accept header for the current request and see if it
       # matches any of the common MIME Types (see Hanami::Action::Mime#MIME_TYPES).
       #
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -181,8 +181,10 @@ module Hanami
       #
       # It prefers, in order:
       #   * Explicit set value (see #format=)
-      #   * Weighted value from Accept
-      #   * Default content type
+      #   * Weighted value from Accept header based on common MIME Types
+      #   * Value from Accept header based on custom registered MIME Types (see Hanami::Controller::Configuration#format)
+      #   * Configured default content type (see Hanami::Controller::Configuration#default_response_format)
+      #   * Hard-coded default content type (see Hanami::Action::Mime::DEFAULT_CONTENT_TYPE)
       #
       # To override the value, use <tt>#format=</tt>
       #
@@ -194,6 +196,8 @@ module Hanami
       # @see Hanami::Configuration#default_request_format
       # @see Hanami::Action::Mime#default_content_type
       # @see Hanami::Action::Mime#DEFAULT_CONTENT_TYPE
+      # @see Hanami::Controller::Configuration#format
+      # @see Hanami::Controller::Configuration#default_response_format
       #
       # @example
       #   require 'hanami/controller'
@@ -207,7 +211,14 @@ module Hanami
       #     end
       #   end
       def content_type
-        @content_type || default_response_type || accepts || default_content_type || DEFAULT_CONTENT_TYPE
+        return @content_type if @content_type
+
+        if accept_header?
+          type = get_common_mime_type_from_accept_header || get_custom_mime_type_from_accept_header
+          return type if type
+        end
+
+        default_response_type || default_content_type || DEFAULT_CONTENT_TYPE
       end
 
       # Action charset setter, receives new charset value
@@ -458,8 +469,8 @@ module Hanami
       #
       # @api private
       def get_common_mime_type_from_accept_header
-          best_q_match(accept, MIME_TYPES)
-        end
+        best_q_match(accept, MIME_TYPES)
+      end
 
       # Look at the Accept header for the current request and see if it
       # matches any of the custom registered MIME Types
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -449,7 +449,8 @@ module Hanami
 
       # Checks if there is an Accept header for the current request.
       #
-      # @return [Boolean] true if there's an Accept header to look at
+      # @return [Boolean] True if there is an Accept header in the current
+      #   request.
       #
       # @since 0.7.0
       # @api private
@@ -458,12 +459,12 @@ module Hanami
       end
 
       # Look at the Accept header for the current request and see if it
-      # matches any of the common MIME Types (see Hanami::Action::Mime#MIME_TYPES)
+      # matches any of the common MIME types (see Hanami::Action::Mime#MIME_TYPES)
       # or the custom registered ones (see Hanami::Controller::Configuration#format).
       #
-      # @return [Nil] when the Accept header does not match any known MIME
-      # Types.
-      # @return [String] the matched MIME type for the given Accept header.
+      # @return [Nil] When the Accept header does not match any known MIME
+      #   type.
+      # @return [String] The matched MIME type for the given Accept header.
       #
       # @since 0.7.0
       #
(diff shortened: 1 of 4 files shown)


russCloak:
I would add more coverage, but there isn't really any isolated/unit tests as is - I'm not sure where they'd be incorporated.

Also, the integration tests I've added are more than enough for the scenario - I suppose I could add more? 

Not sure what the requirements are, please advise.

cyphactor:
First off let me say this awesome. I see this being extremely valuable.

lib/hanami/action/mime.rb
```diff
@@ -207,7 +211,14 @@ def format
       #     end
       #   end
       def content_type
```
cyphactor: Thinking about this purely conceptually the `Content-Type` and response body format should be set based on a couple things according to the `Accepts` header section in the HTTP RFC.

## Scenarios

1. If the service supports one of the mime types specified in `Accepts` header it should respond with the most preferred based on the q scoring that also happens to be supported. See [RFC 2616](http://www.rfc-base.org/txt/rfc-2616.txt) Section 14.1 for details.
2. If the client specifies one or more mime types and the service does not support any of them then the service should respond with a 406 status code.
3. If the client does not specify any accept mime types then the service should default to its ideal mime type.

If I remember correctly the 406 status code is already handled for scenario 2 above, or at least partially handled. Maybe we should revisit that to make sure it is completely handled.

I am a little confused by the priority order of `get_common_mime_type_from_accept_header` and `get_custom_mime_type_from_accept_header`.

It seems in my mind based on the RFC if an accept header is provided with one or more acceptable mime types you have two possible scenarios.

1. the service doesn't support any of them and a 406 status code is responded. This already happens.
2. the service has been identified as supporting one or more of the mime types specified in the `Accepts` header and therefore needs to identified the out of the supported ones which is the most preferred based on the q score stuff and set the content type type to that.

I am not sure why we need `get_common_mime_type_from_accept_header` used at all when we know an accept header has been specified so we have to be in one of the above two states. Am I missing something?

It also seems instead of the `get_custom_mime_type_from_accept_header`, or maybe not instead, but maybe the inside of that method needs to be different. It seems it would need to be something like `best_q_match(accept, get_acceptable_mime_types(accept))`. That way you would be first getting the matching mime types specified by the client in the Accept header and then taking that resulting list and picking the most preferred one using the q score matching.

russCloak: Great feedback @cyphactor, I'm going to make some modifications based on this today.

russCloak:
@cyphactor I've made some adjustments, and added a few more tests to prove the changes work. Can you please take another look?

test/configuration_test.rb
```diff
@@ -187,6 +187,17 @@ def hash
     end
   end
 
+  describe '#format_mime_types' do
+    before do
+      @configuration.format custom: 'custom/format'
+    end
+
+    it 'returns all known format MIME types' do
```
cyphactor: This description should probably read "return all configured format MIME types".

lib/hanami/action/mime.rb
cyphactor: Looking at the changes you made. It looks like it should be covering the RFC now.

test/configuration_test.rb
russCloak: Good thought, I've added this change in 9bc4e2e

cyphactor:
Everything looks good to me at this point. I give it a 👍 .

lib/hanami/action/mime.rb
```diff
@@ -435,12 +447,33 @@ def accept
...
+      # @return [String] the matched MIME type for the given Accept header.
+      #
+      # @since 0.7.0
+      #
+      # @see Hanami::Action::Mime#MIME_TYPES
+      # @see Hanami::Controller::Configuration#format
+      #
+      # @api private
+      def get_mime_type_from_accept_header
+        all_types = (MIME_TYPES + configuration.format_mime_types).uniq
```
cllns: Not sure if it's worth it, but rather than running `uniq` on a new Array, would it be worth trying out `Set` (probably `SortedSet` in this case) and seeing how that performs in comparison?

russCloak: I did a comparison with a few different ways to make a list unique, though there's a difference it's relatively minor. 

```
Running benchmark with array1(10000 elements, 3675 dupes) and array2(10000 elements, 3642 dupes)
                  user       system     total      real
Add Arrays        0.000000   0.000000   0.000000   (0.006018)
New Set (merge)   0.010000   0.000000   0.010000   (0.004714)
New Set with add  0.000000   0.000000   0.000000   (0.005160)
Hash, keys only   0.010000   0.000000   0.010000   (0.004712)
```

With that said, I don't mind switching if it's a real concern.

cyphactor: @cllns what do you think?

beauby: The difference looks pretty insignificant, especially considering that there will usually be around 10 elements max in the `Accept` header, never 1000.

cyphactor: My vote is for simple understandable code in this case rather than a small performance increase.

lib/hanami/action/mime.rb
```diff
@@ -207,7 +212,14 @@ def format
       #     end
       #   end
       def content_type
-        @content_type || default_response_type || accepts || default_content_type || DEFAULT_CONTENT_TYPE
+        return @content_type if @content_type
+
+        if accept_header?
+          type = get_mime_type_from_accept_header
```
beauby: Also (not related to this PR), is there a size limit on the `Accept` header at some level, or can we simply make the CPU spin by supplying a very long list of random values?

lib/hanami/action/mime.rb
beauby: One other thing to keep in mind is that the `configuration.format_mime_types` may typically contain very few values, in which case it is probably less performant to ensure uniqueness than to simply concatenate the arrays (with duplicates) and search in that.

beauby: Also (not related to this PR), is there a size limit on the `Accept` header at some level, or can we simply make the CPU spin by supplying a very long list of random values?

lib/hanami/action/mime.rb
```diff
@@ -435,12 +447,33 @@ def accept
...
+      def accept_header?
+        accept != DEFAULT_ACCEPT
+      end
+
+      # Look at the Accept header for the current request and see if it
+      # matches any of the common MIME Types (see Hanami::Action::Mime#MIME_TYPES)
+      # or the custom registered ones (see Hanami::Controller::Configuration#format).
+      #
+      # @return [Nil] when the Accept header does not match any known MIME
+      # Types.
```
beauby: Nitpicking
> ~~w~~**W**hen the Accept header does not match any known MIME ~~Types~~ **type**.

lib/hanami/action/mime.rb
```diff
@@ -435,12 +447,33 @@ def accept
...
+      # @return [String] the matched MIME type for the given Accept header.
```
beauby: Nitpicking
> ~~t~~**T**he matched MIME type for the given Accept header.

test/integration/mime_type_test.rb
```diff
@@ -162,6 +174,24 @@ def call(params)
     end
 
     it 'sets "Content-Type" header according to "Accept"' do
+      response = @app.get('/custom_from_accept', 'HTTP_ACCEPT' => 'application/custom')
+      response.headers['Content-Type'].must_equal 'application/custom; charset=utf-8'
+      response.body.must_equal                    'custom'
+    end
+
+    it 'sets "Content-Type" header according to "Accept"' do
+      response = @app.get('/custom_from_accept', 'HTTP_ACCEPT' => 'application/custom;q=0.9, application/json;q=0.5')
```
beauby: I've noticed that there does not seem to be any enforced constraint on line size in Hanami but it does improve readability when lines do not exceed 80 columns (or 90, or 100 – 80 just seems to be the inherited standard from the physical terminal days).

lib/hanami/action/mime.rb
```diff
@@ -435,12 +447,33 @@ def accept
...
-      def accepts
-        unless accept == DEFAULT_ACCEPT
-          best_q_match(accept, MIME_TYPES)
-        end
+      def accept_header?
+        accept != DEFAULT_ACCEPT
+      end
+
+      # Look at the Accept header for the current request and see if it
+      # matches any of the common MIME Types (see Hanami::Action::Mime#MIME_TYPES)
```
beauby: Nitpicking:
> MIME ~~T~~**t**ypes

test/integration/mime_type_test.rb
russCloak: I was actually wondering about this, since it seemed like the lines generally are shorter, but definitely not always.

Based on what you said, it seems like the standard is to just make sure the lines are as readable as possible... I can run through all my changes and see where they can be shortened to make it better.

lib/hanami/action/mime.rb
```diff
@@ -435,12 +447,33 @@ def accept
         @accept ||= @_env[HTTP_ACCEPT] || DEFAULT_ACCEPT
       end
 
-      # @since 0.1.0
+      # Checks if there is an Accept header for the current request.
+      #
+      # @return [Boolean] true if there's an Accept header to look at
```
russCloak: Fair enough, I don't mind making these changes - just a note though, I did it this was based on other documentation in the same file (https://github.com/Acornsgrow/controller/blame/resolve_content_type_from_custom_formats/lib/hanami/action/mime.rb#L409). 

Should I clean up *all* of the documentation in the file to match?

russCloak:
@beauby thank you again for the review - I've made all of the changes you suggested. Would you please take another look and see if there's anything else?

lib/hanami/action/mime.rb
beauby: I don't think it's worth the effort but let's see what the maintainers think.

beauby:
As far as I'm concerned it looks good :+1:

russCloak:
@cllns There have been some changes added since you last looked - would you mind giving it a once-over?

lib/hanami/action/mime.rb
cllns: Thanks for looking into this. [`benchmark-ips`](https://github.com/evanphx/benchmark-ips) is a better way of comparing, IMO since it digests the data for you. 

I'm generally in favor of understandability over minor performance gains, but this code would affect every request (as I understand it), where even small gains add up. 

I'll let @jodosha or @joneslee85 determine whether they think this is OK :)

cllns:
Seems fine to me, but I can't merge on this repo 😃 . @jodosha is busy with `rom` integeration for `hanami-model`. @joneslee85 can you take a look?

russCloak:
@cllns and @joneslee85 - any chance we could move forward with this PR? 

I'm going to be using it either way, and I prefer that my company does not need to use only our fork of `hanami/controller` in order to continue using it. 

Thanks in advance!

p.s. Please let me know if there's anything I can do to move this forward.

cllns:
Deferring to @jodosha :)

russCloak:
@jodosha or @joneslee85: Is there no way to get some motion on this PR? 😢

russCloak:
Well, bummer. I'm going to close this PR since I can't get any momentum toward merging or changing it.

It's a shame, I do wish there were more active core members that could review it. In any case, the branch will persist, and it can be looked up if needed.

For now we'll just switch our company to our own fork and use that moving forward. Thanks to all who helped improve this code!

beauby:
IMO this should be merged (whether in its current form or not is up to the maintainers, but the underlying issue is a bug).

cllns:
I understand you're frustrated at the lack of attention, but the core team members are all volunteers working in their spare time. 

I expect this will be merged and you can get back over to using regular ole' `hanami-controller` soon :)

I do recommend, if you want this PR to get merged, that you should leave it open. I imagine it'll get overlooked if it's closed. 🌸

russCloak:
@cllns Fair enough - I'll reopen. I just figured a stale PR is just as bad as a closed one.

Ideally it will get in eventually, and we'll head back over to this remote - thanks for the reply!

jodosha:
@russCloak I'm sorry if we're merging only today.

All of you @russCloak, @cyphactor, @beauby, and @cllns and it would be a shame that good collaboration like this would be lost. 💯 

Please understand that we all do voluntary work and other stuff can take higher priority. For instance other parts of Hanami, our day jobs, our families.. It takes time the right to digest a month of discussions, to check the code, tweak it if needed. That's the reasons for the waiting. 🌸 

That being said, thank you for this contribution. This is a great fix. 💚

PR #187, opened 2016-11-23
Fix references to the old `Hanami::Repository` API.

beauby:
Also, line 989 seems to be missing a *not*.


AlfonsoUceda:
@beauby thanks!

PR #188, opened 2016-11-23
Wrapper to allow consistent session hash access

### LINKED ISSUE #185: [Discussion] Is `session` supposed to return an indifferent hash?
My colleague @createdbypete's just pointed out for me there is a regression with the accessor `action.session` when he upgraded our internal app to Hanami 0.9

Before:

```
action.session[:key] #=> the value
```

Now:

```
action.session[:key] #=> nil
```

I could not find any traces in the session code that cast to `Utils::Hash` so I assume that it is with intention `session` would return a original hash or whatever that `env` returns.

Now let's take a moment to discuss if `session` should be casted to `Utils::Hash` OR respect whatever rack returns. There is already a similar pattern with `params` in which we cast and for the `raw.params` we leave it intact. 

Thoughts? cc @hanami/core-team @createdbypete

createdbypete:
Rack always [converts session keys to a string](https://github.com/rack/rack/blob/1-6-stable/lib/rack/session/abstract/id.rb#L23), however in Hanami Action unit tests the middleware is not invoked and the Rack `SessionHash` is not used and a plain Hash is passed in. 

This provided inconsistent accessor behaviour as Hanami prefers symbolic keys be used for access and Ruby's default `Hash` obviously does not convert the keys to strings by default.

The proposed new class `Hanami::Action::Session::SessionHash` attempts to intercept the accessor methods where the type conversion is important to ensure that occurs and delegate everything else to the original object (either the Rack `SessionHash` or the Ruby `Hash`).

Closes #185

### DIFF
--- a/lib/hanami/action/session.rb
+++ b/lib/hanami/action/session.rb
@@ -1,4 +1,5 @@
 require 'hanami/action/flash'
+require 'hanami/action/session/session_hash'
 
 module Hanami
   module Action
@@ -58,7 +59,7 @@ module Hanami
       #     end
       #   end
       def session
-        @_env[SESSION_KEY] ||= {}
+        SessionHash.new(@_env[SESSION_KEY] || {})
       end
 
       # Read errors from flash or delegate to the superclass
--- a/lib/hanami/action/session/session_hash.rb
+++ b/lib/hanami/action/session/session_hash.rb
@@ -1,0 +1,30 @@
+module Hanami
+  module Action
+    module Session
+      # Wrapper around session hash to provide consistent key access.
+      #
+      # Since 0.8.0
+      class SessionHash < ::SimpleDelegator
+        def [](key)
+          __getobj__[key.to_s]
+        end
+        alias :fetch :[]
+
+        def has_key?(key)
+          __getobj__.has_key?(key.to_s)
+        end
+        alias :key? :has_key?
+        alias :include? :has_key?
+
+        def []=(key, value)
+          __getobj__[key.to_s] = value
+        end
+        alias :store :[]=
+
+        def delete(key)
+          __getobj__.delete(key.to_s)
+        end
+      end
+    end
+  end
+end
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -22,5 +22,12 @@ describe Hanami::Action do
 
       action.exposures[:session].must_equal(session)
     end
+
+    it 'allows value access via symbolic keys' do
+      action = SessionAction.new
+      action.call({'rack.session' => { 'foo' => 'bar' }})
+
+      action.session[:foo].must_equal('bar')
+    end
   end
 end


beauby:
Wait, if outside of tests, rack allows indifferent access (by converting all keys to strings under the hood), and during unit tests, rack does nothing, then providing a hash with symbol keys should fix your issue, no?

beauby:
Could you show a controller + unit test where this issue arises? I'd like to understand.

createdbypete:
@beauby that's a good point. In my tests I'd been trying to keep the same types as rack would user but perhaps given the conversion it would do anyway this doesn't actually matter and using symbols in my test is the best approach.

beauby:
Basically in unit tests you have to provide everything in the state (better yet in a compatible interface) you would expect it to be in the controller.

createdbypete:
@beauby thanks, this has probably lead to my confusion. Was just tracking down why I thought to use the strings in the first place, and I believe this example was the cause: https://github.com/hanami/controller#sessions

```ruby
action = ReadSessionFromRackEnv.new
action.call({ 'rack.session' => { 'age' => '31' }})
action.session[:age]
#=> 31
```

createdbypete:
Perhaps a better PR (since seems this should be closed?) would be to update the guides and tests to be consistent with using symbols incase others are wrongly led by example?

- https://github.com/hanami/controller/blob/master/test/session_test.rb (use symbols)
- https://github.com/hanami/controller/blob/master/README.md#sessions (use symbols)
- https://github.com/hanami/hanami.github.io/blob/build/source/guides/actions/sessions.md (add heading about how to test?)

What do you think? Also if you can think of any other places?

joneslee85:
LGTM

cc @hanami/core-team

createdbypete:
@beauby / anyone else. More of a style thing but perhaps something to consider is providing symbols in the tests will mean using mixed key type in the hash. Example:

```ruby
params = {'rack.session' => {:foo => 'bar'}, :baz => '123'}
action.call(params)
```

Where `rack.session` obviously needs to be a string key (and any other rack specific ones) whereas everything else goes in as a symbol. Nothing I'm going to lose sleep over but might be something Hanami cares about.

jodosha:
@createdbypete At the beginning I didn't understood that you were talking about unit tests. This changes everything.

If you look at [this commit](https://github.com/hanami/controller/commit/87686db5eff879e67b5d8acc97d58ae99e3b544f), we ensure that symbol access is working with a real Rack env.

---

Now, I think that an interface is weak if relies too much on the documentation. In our case, we want to make sure that devs can simulate Rack session in any from: both with string and/or symbol keys.

If the scope of this PR is to support devs in this process we should:

  * Keep the unit test of this PR, but change its description, to specify symbol access for testing.
  * Revert all the other code
  * Ensure that params/env in testing mode will return symbolized hash. See: https://github.com/hanami/controller/blob/87686db5eff879e67b5d8acc97d58ae99e3b544f/lib/hanami/action/base_params.rb#L146
  * Rebase in a single commit

WDYT?

/cc @joneslee85 @beauby

beauby:
@createdbypete @jodosha I agree that the current situation is not ideal as it breaks POLS in a way, but I think we should be careful to introduce as few proxy objects as possible (especially when they end up being proxy objects for proxy objects, like here).

beauby:
One possible solution to most of those problems would be to introduce some helper class `Hanami::RackEnvBuilder` that would be used in test to setup this stuff transparently. It's not perfect either but it avoids adding code for testing purposes at the expense of production code.

createdbypete:
I'm not sure how the `BaseParams` comes into this? Happy to make whatever changes but I can't see how that class will make a difference with the session key access?

beauby:
@createdbypete Because the controller offers a unified access to parameters (be it session, query parameters, router parameters, post parameters, etc.), so consistent behavior between session and base params (in particular) should be ensured.

mereghost:
@createdbypete Any progress on this issue?

jodosha:
Closing in favor of #203

PR #189, opened 2016-11-27
Update hanami utils and use #deep_symbolize!

AlfonsoUceda:
Related to https://github.com/hanami/utils/pull/162

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ unless ENV['TRAVIS']
 end
 
 gem 'minitest',      '~> 5.8'
-gem 'hanami-utils',  '~> 0.9', require: false, github: 'hanami/utils',  branch: '0.9.x'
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
cc @alexd16

beauby:
For the record, what's the rationale behind symbolizing everything? In every case I can think of, it feels like a waste of CPU for little gains in code clarity.

AlfonsoUceda:
cc @jodosha

cllns:
@beauby I like symbolizing everything. It's almost like having indifferent access, without all the complexity. It provides a more consistent interface, letting you abstract over the implementation detail of whether the input hash uses strings or symbols for keys.

Also, I think the CPU cycles are negligible? I could be wrong though.

jodosha:
@beauby It's true, the implementation is not efficient.

But each design comes with a tradeoff. In this case what we value the most is the API: provide symbol-only access. Then we can work to make it to perform better.

PR #191, opened 2016-12-05
MIME Type to format resolution

mereghost:
Introduces Hanami::Action::Mime::MIME_TYPES with the most common (realistic) MIME types used on web applications and their most used file extensions.

Solves hanami/hanami#663

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -41,6 +41,64 @@ module Hanami
       # @api private
       DEFAULT_CHARSET = 'utf-8'.freeze
 
+      # Most commom mime types used for responses
+      #
+      # @since 0.9.0
+      # @api private
+      MIME_TYPES = {
+        txt: 'text/plain',
+        html: 'text/html',
+        json: 'application/json',
+        manifest: 'text/cache-manifest',
+        atom: 'application/atom+xml',
+        avi: 'video/x-msvideo',
+        bmp: 'image/bmp',
+        bz: 'application/x-bzip',
+        bz2: 'application/x-bzip2',
+        chm: 'application/vnd.ms-htmlhelp',
+        css: 'text/css',
+        csv: 'text/csv',
+        flv: 'video/x-flv',
+        gif: 'image/gif',
+        gz: 'application/x-gzip',
+        h264: 'video/h264',
+        ico: 'image/vnd.microsoft.icon',
+        ics: 'text/calendar',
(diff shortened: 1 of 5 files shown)


mereghost:
Hmmm... seems that some update to dry-validation broke the build. Investigating.

mereghost:
Seems that `dry-validation` changed behaviour as evidenced by [this commit](https://github.com/dry-rb/dry-validation/commit/dfddad5010fa3df4d129cdd87e70dbdc3c27b2b3).

I will make the necessary changes in the tests to adequate to the new behaviour.

lib/hanami/action/mime.rb
```diff
@@ -41,6 +41,64 @@ module Mime
       # @api private
       DEFAULT_CHARSET = 'utf-8'.freeze
 
+      # Most commom mime types used for responses
+      #
+      # @since 0.9.0
+      # @api private
+      MIME_TYPES = {
```
mereghost: Makes sense for me.

lib/hanami/controller/configuration.rb
```diff
@@ -416,7 +416,7 @@ def format(hash)
       def mime_types
         @mime_types ||= begin
                           ((@formats.keys - DEFAULT_FORMATS.keys) +
-                          ::Rack::Mime::MIME_TYPES.values).freeze
+                          Hanami::Action::Mime::MIME_TYPES.values).freeze
```
mereghost: My bad, I totally forgot about that. Any suggestions on what to move to the top of the list? My bet would be `html, txt, json, pdf, zip` in that order.

jodosha: @mereghost If you agree, we can do `txt`, `html`, `json` in this order, and leave the rest as it is.

lib/hanami/action/mime.rb
jodosha: > Yeah, I tried to avoid the multiple file extension trouble

Of course, the problem of hanami/hanami#663 is caused by duplicates. If we introduce them again, we haven't solved too much 😉 .

jodosha:
**Please don't merge yet.**

mereghost:
Want me to squash it to a single commit?

jodosha:
@mereghost Yes, thanks 👍

mereghost:
@jodosha should we merge this?

jodosha:
@mereghost Yes, it is the time to :shipit: :wink:

PR #193, opened 2016-12-07
Don't annotate exception in rack.env when exception handler handled the exception.

thhermansen:
We had defined in our Hanami app a handler for an exception class. A subclass of that exception was raised and the handler was correctly called, but the error was still logged to sentry (which integrates with rack [here](https://github.com/getsentry/raven-ruby/blob/5414217bfe0273ede358319503ae4a71db45be97/lib/raven/integrations/rack.rb#L58-L59).)

It seems to me that the exception was set on rack's env even when it was handled due to the fact that the `#handled_exception?` didn't check for subclasses as the `#exception_handler` did.

Fixed it by using the same code to figure out if we have a handler for exception class or not.

### DIFF
--- a/lib/hanami/controller/configuration.rb
+++ b/lib/hanami/controller/configuration.rb
@@ -192,16 +192,7 @@ module Hanami
       #
       # @see Hanami::Controller::Configuration#handle_exception
       def exception_handler(exception)
-        handler = nil
-
-        @handled_exceptions.each do |exception_class, h|
-          if exception.kind_of?(exception_class)
-            handler = h
-            break
-          end
-        end
-
-        handler || DEFAULT_ERROR_CODE
+        find_exception_handler(exception) || DEFAULT_ERROR_CODE
       end
 
       # Check if the given exception is handled.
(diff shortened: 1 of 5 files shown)


AlfonsoUceda:
@thhermansen it works perfect, could you squash into one commit? thank!

cc @jodosha

thhermansen:
Thank you for taking a look at this. I have squashed it in to one commit and replaced this pull request with the squashed version.

jodosha:
@thhermansen Thanks for this PR.

I'm struggling to find where is the test that checks if `rack.exception` isn't in Rack env.

thhermansen:
@jodosha - I think you are right. I wrote a test in `rack_errors_test.rb` which failed first and passed after changing implementation in configuration. I did not notice the direct test on `rack_exception_test.rb` until now. I have added a test there which directly asserts `env['rack.exception']` to be nil when subclass of an exception is handled.

jodosha:
@thhermansen I've just merged this, thank you! 👍

thhermansen:
Thank you! :-)

PR #195, opened 2016-12-15
Expose flash by default

jodosha:
Reported by @nsheremet in chat.

There is a regression that prevents to expose `flash` outside of an action. This PR fixes the problem.

/cc @hanami/issues for review. Thanks 👍

### DIFF
--- a/test/session_test.rb
+++ b/test/session_test.rb
@@ -23,4 +23,16 @@ describe Hanami::Action do
       action.exposures[:session].must_equal(session)
     end
   end
+
+  describe 'flash' do
+    it 'exposes session' do
+      action = FlashAction.new
+      action.call({})
+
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
👍 will you add a test in hanami repo? maybe if we want to check to use in the view.

nsheremet:
👍

test/session_test.rb
```diff
@@ -23,4 +23,16 @@
       action.exposures[:session].must_equal(session)
     end
   end
+
+  describe 'flash' do
+    it 'exposes session' do
```
Neodelf: It is typo? May be correctly `it exposes flash` ?

jodosha: @Neodelf Thanks for catching it. 👍

PR #200, opened 2017-01-08
Return only non-stale flash values

### LINKED ISSUE #196: Flash returns the same message twice
I'm having a weird issue using flash with controller 0.8.1, where I get the same message twice. I'm POSTing to `save.rb`, saving the message there, and then catching the message in `index.rb`:

```
// save.rb
module Web::Controllers::Home
  class Save
    include Web::Action

    def call(params)
      flash[:message] = 'Saved!'
      redirect_to routes.dashboard_path
    end
  end
end
```

```
// index.rb
module Web::Controllers::Home
  class Index
    include Web::Action

    expose :message

    def call(params)
      p "HI THERE -> #{flash[:message]}"
      @message = flash[:message]
    end
  end
end
```

On the first load of index, it prints out `"HI THERE -> Saved!"`
On the _second_ load, it prints out `"HI THERE -> {:message=>\"Saved!\"}"`

Am I doing something wrong? Is this related to #195?

matiasleidemer:
`Hanami::Action::Flash#[]` was taking into consideration stale values,
causing weird behaviors when fetching flash data after refreshing a
redirected page.

This commit fixes this issue by allowing only non-stale flash values to
be taken into consideration.

I'm not 100% sure if this is the right fix or not, since stale data should not be available in the `flash` object in the first place (I believe...). 

Fixes #196

### DIFF
--- a/lib/hanami/action/flash.rb
+++ b/lib/hanami/action/flash.rb
@@ -55,7 +55,7 @@ module Hanami
       # @api private
       def [](key)
         last_request_flash.merge(data).fetch(key) do
-          _values.find {|data| !data[key].nil? }
+          _available_values.find { |data| !data[key].nil? }
         end
       end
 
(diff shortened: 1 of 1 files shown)


matiasleidemer:
Hey @jodosha, thanks for the feedback! Unfortunately things have been a little crazy here and I haven't, yet, taken the time to address your comments. Will do it asap.

jodosha:
Closing in favor of #202

PR #201, opened 2017-01-08
Make hanami-controller Rack 2.0 compatible

davydovanton:
Closes https://github.com/hanami/controller/issues/199

### DIFF
--- a/lib/hanami/action/glue.rb
+++ b/lib/hanami/action/glue.rb
@@ -59,7 +59,7 @@ module Hanami
       #
       # @since 0.4.3
       def sending_file?
-        @_body.is_a?(::Rack::File)
+        @_body&.first.is_a?(::Rack::File)
       end
     end
   end
--- a/lib/hanami/action/rack/file.rb
+++ b/lib/hanami/action/rack/file.rb
@@ -10,20 +10,27 @@ module Hanami
       #
       # @see Hanami::Action::Rack#send_file
       class File
+
+        # The key that returns path info from the Rack env
+        #
+        # @since x.x.x
+        # @api private
+        PATH_INFO = "PATH_INFO".freeze
+
         # @param path [String,Pathname] file path
         #
         # @since 0.4.3
         # @api private
         def initialize(path)
-          @file = ::Rack::File.new(nil)
+          @file = ::Rack::File.new(Dir.pwd)
           @path = path
         end
 
         # @since 0.4.3
         # @api private
         def call(env)
-          @file.path = @path.to_s
-          @file.serving(env)
+          env[PATH_INFO] = @path
+          @file.get(env)
         rescue Errno::ENOENT
           [404, {}, nil]
         end
--- a/hanami-controller.gemspec
+++ b/hanami-controller.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.3.0'
 
-  spec.add_dependency 'rack',         '~> 1.6', '>= 1.6.2'
+  spec.add_dependency 'rack',         '~> 2.0'
   spec.add_dependency 'hanami-utils', '~> 1.0.0.beta1'
 
   spec.add_development_dependency 'bundler',   '~> 1.6'


lib/hanami/action/glue.rb
```diff
@@ -59,7 +59,7 @@ def finish
       #
       # @since 0.4.3
       def sending_file?
-        @_body.is_a?(::Rack::File)
+        @_body.is_a?(::Rack::File) || @_body&.first.is_a?(::Rack::File)
```
davydovanton: Rack 2 replaced `#each` method to `Iterator` class ([link](https://github.com/rack/rack/commit/e0ac32913a725899b3ee6bec9f293d6a01469cae#diff-894804c816a11584eb28499cb8bbe396L99)). That's why now action wrap `Rack::File` to array

lib/hanami/action/rack/file.rb
```diff
@@ -15,15 +15,15 @@ class File
...
+          @file = ::Rack::File.new(Dir.pwd)
           @path = path
         end
 
         # @since 0.4.3
         # @api private
         def call(env)
-          @file.path = @path.to_s
-          @file.serving(env)
+          env["PATH_INFO"] = @path
```
davydovanton: And you can see that now Rack take path from `path_info` env variable ([link](https://github.com/rack/rack/blob/master/lib/rack/file.rb#L40))

lib/hanami/action/glue.rb
istickz: in this place you may get nil after `@_body&.first` and `undefined method `is_a' for nil:NilClass`

Bugagazavr:
```
2.3.0 :001 > nil.is_a?(String)
 => false
```

istickz: You're right! I just forgot `?` in method name. thx

Gemfile
```diff
@@ -8,7 +8,7 @@ end
 
 gem 'minitest',      '~> 5.8'
 gem 'hanami-utils',  '~> 1.0.0.beta1', require: false, github: 'hanami/utils',  branch: '1.0.x'
-gem 'hanami-router', '~> 1.0.0.beta1', require: false, github: 'hanami/router', branch: '1.0.x'
+gem 'hanami-router', '~> 1.0.0.beta1', require: false, github: 'pascalbetz/router', branch: 'issue-136'
```
pascalbetz: I assume this should be pointing to _hanami/router_ before this PR is merged.

davydovanton: added this only for check that all tests in CI will pass :)

davydovanton:
@pascalbetz updated gemfile 👍

lib/hanami/action/glue.rb
jodosha: @davydovanton Because of that change, can we improve this implementation?

For instance:

  * Can we remove the first part of the conditional?
  * If in case of file sending, `@_body` is an instance of `Iterator` can we just do `@_body.kind_of?(Iterator)`?

If those changes aren't possible, please merge it.

lib/hanami/action/rack/file.rb
jodosha: @davydovanton Can you please use a constant for `"PATH_INFO"`? Thanks!

davydovanton:
@jodosha thanks for review 👍 I'll fix `#sending_file?` as soon as possible (1-2 days I think)

lib/hanami/action/glue.rb
davydovanton: I think we can remove the first part of the conditional because rack2 wrap all `Rack::File` instances to array like a normal body

davydovanton:
@jodosha @pascalbetz I fixed all issues. I think we can merge routes and controller PRs now 👍

jodosha:
The CI build was failing because `hanami/router` `1.0.x` branch wasn't already with the Rack 2.0 update. I fixed that and tried this PR locally with success.

As long CI is green again I'm gonna merge it. Thanks @davydovanton !

davydovanton:
All  are green 🎉
Thanks for review @jodosha and @pascalbetz ❤

PR #202, opened 2017-01-21
Ensure flash to return only fresh values

### LINKED ISSUE #196: Flash returns the same message twice

jodosha:
Closes #196
Ref #200

Credit goes to @matiasleidemer for finding the right solution. Thank you Matias!

### DIFF
--- a/lib/hanami/action/flash.rb
+++ b/lib/hanami/action/flash.rb
@@ -142,7 +142,7 @@ module Hanami
       # @since 0.3.0
       # @api private
       def _values
-        flash.values
+        flash.select { |session_id, _| !delete?(session_id) }.values
       end
 
       # Determine if delete data from flash for the given Request ID
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -1274,6 +1274,25 @@ module FullStack
       end
     end
 
+    module Settings
+      class Index
+        include FullStack::Action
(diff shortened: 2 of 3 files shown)


jodosha:
Note: JRuby is broken on Travis CI at this moment. The build passes for MRI 2.3/2.4 on CI and for JRuby on my local machine too. Merging this.

matiasleidemer:
Looks very nice! sorry I didn't help more (with the PR), but I'm glad it fixes 👍

PR #203, opened 2017-01-21
Ensure symbol access for session during action unit tests

### LINKED ISSUE #185: [Discussion] Is `session` supposed to return an indifferent hash?

jodosha:
Based on initial implementation by @createdbypete (see #188).
Closes #185

### DIFF
--- a/lib/hanami/action/base_params.rb
+++ b/lib/hanami/action/base_params.rb
@@ -15,6 +15,18 @@ module Hanami
       # @since 0.7.0
       ROUTER_PARAMS = 'router.params'.freeze
 
+      # The key that returns Rack session params from the Rack env
+      # Please note that this is used only when an action is unit tested.
+      #
+      # @since x.x.x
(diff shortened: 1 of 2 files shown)


jodosha:
At the time I'm writing JRuby is broken on CI. Merging regardless of this problem.

PR #204, opened 2017-01-22
Do not send file outside of public directory

jodosha:
This PR fixes two problems:

  * Don't let files outside of public directory to be sent
  * Fix the `#renderable?` check, which causes a server side error

This completes #201 

---

/cc @davydovanton @hanami/issues for review.


jodosha:
The CI fails on JRuby because Travis CI is having problems since a couple of days.

AlfonsoUceda:
@jodosha 👍

PR #212, opened 2017-02-16
Works well with empty cache

davydovanton:
## Why we need this changes?

I found that some middlewares use strange practice. They set `HTTP_IF_MODIFIED_SINCE` or `HTTP_IF_NONE_MATCH` as empty string.
Simple example: rack-mini-profiler, When I worked with this integration I found [this lines of code](https://github.com/MiniProfiler/rack-mini-profiler/blob/master/lib/mini_profiler/profiler.rb#L251-L254). As you can understand these lines break hanami cache. That's why I think we need to add the check to the empty line too.

WDYT?

/cc @hanami/core @hanami/ecosystem

### DIFF
--- a/lib/hanami/action/cache/conditional_get.rb
+++ b/lib/hanami/action/cache/conditional_get.rb
@@ -56,7 +56,7 @@ module Hanami
         end
 
         def fresh?
-          modified_since && Time.httpdate(modified_since).to_i >= @value.to_i
+          modified_since && !modified_since.empty? && Time.httpdate(modified_since).to_i >= @value.to_i
         end
 
         def header
--- a/lib/hanami/action/cache/conditional_get.rb
+++ b/lib/hanami/action/cache/conditional_get.rb
@@ -56,7 +56,7 @@ module Hanami
         end
 
         def fresh?
-          modified_since && !modified_since.empty? && Time.httpdate(modified_since).to_i >= @value.to_i
+          !Hanami::Utils::Blank.blank?(modified_since) && Time.httpdate(modified_since).to_i >= @value.to_i
         end
 
         def header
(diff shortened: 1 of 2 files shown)


cllns:
could it ever be `'  '` (a string with spaces), or other "blank" values? I don't think so, but if it could, we should consider using [`Hanami::Utils::Blank`](https://github.com/hanami/utils/blob/master/lib/hanami/utils/blank.rb).

This seems good to me, but let's have @jodosha weigh in.

AlfonsoUceda:
@davydovanton could you make a test for `HTTP_IF_NONE_MATCH`? :+1:

davydovanton:
Yes, of course!

davydovanton:
@cllns good idea, I'll fix it today and will add tests :+1:

davydovanton:
@cllns @AlfonsoUceda fixed all issues 👍

davydovanton:
@jodosha could you check this PR too? 😊

davydovanton:
@jodosha Thanks for feedback 👍

I added new tests for `Last-Modifed` header.

When we have **ANY** not nil or false value in `IF_MODIFIED_SINCE` variable we will return time for `Last-Modified`. If `IF_MODIFIED_SINCE` is nil/false and `HTTP_IF_NONE_MATCH` is empty string we will return `nil` for this header.

That's why I have a question: WDYT should we returns nil when `IF_MODIFIED_SINCE` is empty string?

jodosha:
@davydovanton In general, when a header is `nil` we shouldn't send the header at all.

In the tests you wrote:

```ruby
response.headers['Last-Modified'].must_equal nil
```

But it should be:

```ruby
assert !response.headers.key?('Last-Modified')
```

---

The two versions are similar, but not equivalent.

In the first case we are asserting that `'Last-Modified'` returns `nil`, which isn't completely correct. We want to assert that the header is not set at all.

Unfortunately, Ruby is misleading when it comes to `nil` values.

```ruby
headers = { "Last-Modified" => nil }
headers['Last-Modified'].must_equal nil # It passes, but it's not the end result that we want
```

```ruby
headers = { "Last-Modified" => nil }
assert !response.headers.key?('Last-Modified') # It fails because the key it's set
```

---

I'm gonna merge and adjust accordingly.

PR #215, opened 2017-02-23
Keep flash after redirect

### LINKED ISSUE #213: Regression: flash after redirect broken
Flash messages should survive after a redirect. They used to with `0.9.2`, but now they don't, with `1.0.0.beta1`. 

This is a result from #202. Specifically, [the change to the implementation of `_values` in `lib/hanami/action/flash.rb`](https://github.com/hanami/controller/pull/202/files#diff-ce13f23d4ed57d06a91f48c61b42c054L145). I reverted that one line locally and flashes survive redirects.

Note: that line should not just be reverted, since we should still address #196.

Discovered in parallel (and simultaneously!) by @deangerber. I think someone else mentioned this recently too, but I can't find who it was.

jodosha:
Closes #213

### DIFF
--- a/test/integration/full_stack_test.rb
+++ b/test/integration/full_stack_test.rb
@@ -52,11 +52,11 @@ describe 'Full stack application' do
     post '/settings', {}
     follow_redirect!
 
-    last_response.headers['X-Flash'].must_equal 'Saved!'
+    last_response.body.must_match %r{Hanami::Action::Flash:0x[\d\w]* {:message=>"Saved!"}}
 
     get '/settings'
 
-    last_response.headers['X-Flash'].must_equal "Blank message"
+    last_response.body.must_match %r{Hanami::Action::Flash:0x[\d\w]* {}}
   end
 
   it 'can access params with string symbols or methods' do
--- a/lib/hanami/action/flash.rb
+++ b/lib/hanami/action/flash.rb
@@ -31,6 +31,7 @@ module Hanami
         @session         = session
         @request_id      = request_id
         @last_request_id = session[LAST_REQUEST_KEY]
+        @merged          = {}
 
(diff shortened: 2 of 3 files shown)


jodosha:
@cllns Please review, I updated the tests according to the strategy that we just discussed about.

test/integration/full_stack_test.rb
```diff
@@ -52,11 +52,11 @@ def app
     post '/settings', {}
     follow_redirect!
 
-    last_response.headers['X-Flash'].must_equal 'Saved!'
+    last_response.body.must_match %r{Hanami::Action::Flash:0x[\d\w]* {:message=>"Saved!"}}
```
cllns: I think we can do `\h{12}` instead of `[\d\w]*` ? or at least `\h+`

test/integration/full_stack_test.rb
```diff
@@ -52,11 +52,11 @@ def app
...
 
     get '/settings'
 
-    last_response.headers['X-Flash'].must_equal "Blank message"
+    last_response.body.must_match %r{Hanami::Action::Flash:0x[\d\w]* {}}
```
cllns: same as above

jodosha:
This number depends on the architecture where the Ruby VM is running on. I prefer a relaxed regexp, cause it isn't too much important here.

http://lucaguidi.com

cllns:
hm, I see :) well `\w` contains `\d` (unless that's not true on some platforms too 😝 )

AlfonsoUceda:
👍

jodosha:
@cllns 
> hm, I see :) well \w contains \d (unless that's not true on some platforms too 😝 )

It's about the length of the string, rather than the contents.

PR #216, opened 2017-02-23
Introduces #unsafe_send_file

### LINKED ISSUE #210: new #send_file behavior is error-prone in globbed routes
version: 1.0.0-beta1 

https://github.com/hanami/controller/commit/4a1e1af2c6da5b6adf8d9d3c17b397dbd3487b5e restricts send_file to files only in public_directory, and allows the router to attempt other routes.  This causes an nasty problem that isn't obvious.

When using a glob route, say:
```
  get '/path/:id(.:format), to 'example#show'
```
and the `#show` action does a `#send_file` that isn't in the public folder, the return value is has a "X-Cascade"=>"pass" header.  This entails a second attempt on the same route, except :format isn't globbed out (:id is now "#{id}.#{format}").  This will almost certainly fail with a 404/406 on a well written controller, and the original "File Not Found" 404 error message is lost to the new "ID is invalid".  Looking for a previous failure after handing the controller a new set of (invalid) params doesn't seem like a practice I want in my controllers.

Moreover, a controller that authoritatively says "send this file as a body" shouldn't fail silently - it should be a hard 404 or 500 if unable to do so (in [some|many|most] cases), just like a missing template.   If people also need a soft-failing `#send_file`, perhaps the weaker version should be `#attempt_send_file`, or alternatively move the more firm version to `#send_file!`, so we don't have failures coming from other controllers or another pass through the same controller with bogus data.

### LINKED ISSUE #211: need an [unsafe] method to #send_file on a location outside of the public_directory
version: 1.0.0.beta-1

https://github.com/hanami/controller/commit/4a1e1af2c6da5b6adf8d9d3c17b397dbd3487b5e restricts #send_file to the public_directory.  While that's a well-intentioned security measure, it's a common use-case to `#send_file` CMS content, or things generated dynamically via `TempFile`.  I don't think Hanami needs the complexity of managing whitelist-able folders, but I do need to be able to do my own permission checking and still use `#send_file`.

a `#send_file_unsafe` (or some better name) should be added, and noted in the CHANGELOG that you may need to use this in lieu of the current `#send_file` behavior.

It might be worth putting a deprecation note in the current (hanami 0.9.X / controller 0.8.X) version if you're going to make existing behavior less accommodating, and fail silently after upgrading (https://github.com/hanami/controller/issues/210)?

mereghost:
- We still provide send_file as a safe alternative, only sending files
located on the public directory
- `#unsafe_send_file` allows the user to send a file from anywhere while
reminding him that this might be risky.

- Fixes #211

### DIFF
--- a/lib/hanami/action/rack.rb
+++ b/lib/hanami/action/rack.rb
@@ -290,8 +290,6 @@ module Hanami
 
       # Send a file as response from anywhere in the file system.
       #
-      # It automatically handle the following cases:
-      #
       # @see Hanami::Action::Rack#send_file
       #
       # @param path [String, Pathname] path to the file to be sent
--- a/lib/hanami/action/rack.rb
+++ b/lib/hanami/action/rack.rb
@@ -257,6 +257,7 @@ module Hanami
       end
 
       # Send a file as response.
+      #  <tt>This method only sends files from the public directory</tt>
       #
       # It automatically handle the following cases:
       #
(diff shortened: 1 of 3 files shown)


lib/hanami/action/rack.rb
```diff
@@ -287,6 +288,32 @@ def send_file(path)
         halt result[0], result[2]
       end
 
+      # Send a file as response from anywhere in the file system.
```
cllns: Can we add a note here explaining why this is dangerous? Perhaps in bold? :)

tercenya:
this is exactly what https://github.com/hanami/controller/issues/211 was asking for, but I don't think it addresses https://github.com/hanami/controller/issues/210 at all.

Let me see about mustering a more complete example or demonstrating some failing tests, as to show why the problem with 210 is insidious.

tercenya:
The merits of whether you think this sufficient resolution to #210 are over there, as not to clutter up the pull request.  If you disagree, another solution would be to document the new `#send_file` and `#unsafe_send_file` in the UPGRADE guide, alongside this PR.

jodosha:
I agree, this solves only #211, but not #210. I edited the PR description accordingly.

PR #217, opened 2017-02-27
Send file with formatting globs

jodosha:
Ref https://github.com/hanami/controller/issues/210


jodosha:
Closing in favor of #218

PR #229, opened 2017-06-13
Fix the parsing and evaluation of the Accepts Header

### LINKED ISSUE #225: Detected MIME/format changed from 0.9 to 1.0
During migration to Hanami 1.0 (from 0.9), I've noticed that now my `json` actions require to specify `self.format = :json` or response will be returned as `html` otherwise.

Accordingly to gitter chat with @jodosha , it seems that bug could be introduced in #206 or #191

mereghost:
# Accept Header Parsing

First of all the [Accept Header specification](https://tools.ietf.org/html/rfc7231#section-5.3.2) mentions that any media range has lower priority than the more specific ones, so `text/html` has higher precedence than `text/*` and `*/*` even at the same `q` (quality) value.

The order of the media types seem to be irrelevant and in case of ties the server could return any of the requested media ranges.

## Our Troubles

As evidenced by [hanami/controller#225](https://github.com/hanami/controller/issues/225) something seems broken. At first I thought that this line of code was the problem: 

```ruby
if req_mime == DEFAULT_ACCEPT
  # See https://github.com/hanami/controller/issues/167
  match = default_content_type
else
```

What this does is: if the mime type being checked is `*/*`  we should return the configured default content type (`text/html` in hanami's case, `application/octet-stream` on controller' s).

That led me to this part of the code:

```ruby
values = ::Rack::Utils.q_values(q_value_header)
values = values.map do |req_mime, quality|
  if req_mime == DEFAULT_ACCEPT
    # See https://github.com/hanami/controller/issues/167
    match = default_content_type
  else
    match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
  end
  next unless match
  [match, quality]
end.compact
```

Here's the bigger picture and what pointed me to even bigger problems. `Rack::Mime.match?` tries to *expand* the media ranges, so it will try to match `text/*` to `text/plain` for example.

The method from where this was taken has the following signature: `best_q_match(req_mimes, available_mimes)`. If we call it with a `req_mimes` of `application/json,text/html,*/*,text/*` what this snippet will output is: `[['application/json', 1.0], ['text/html', 1.0], ['text/html', 1.0], ['text/plain', 1.0]]`.

Then we proceed to sort this matrix using a custom sorting algorithm (same code is found on `rack 2.0`):

```ruby
 value = values.sort_by do |match, quality|
   (match.split('/'.freeze, 2).count('*'.freeze) * -10) + quality
 end.last
```

By then, no media ranges reach this part of the code, so the priority rule never gets applied. So we end up with `text/plain` when we should have returned either `text/html` or `application/json`.

### Differences when the Action has an `accept :json`

With the same `req_mime` as before and an `available_mimes` of `application/json` (the side effect of sending `Action.accept`) the result would be `text/html`. Why? Because the check for the `DEFAULT_ACCEPT` ignores the `available_mimes`, it ends up in the result, gets sorted and picked in the end. 

## Solution

I borrowed a lot of ideas on how to solve it by looking at [negotiator](https://github.com/jshttp/negotiator) and seeing how the python, nodejs communities (among others) solved this.

I created an object (name `Specification` which is a crappy name, so I'm open to suggestions) that encapsulates the mime specification, its quality, index on the accepts header and the format solved by `Rack::Mime.match`.

Said object then calculates the priority and in case of ties opts for the type nearest to the start of the header.

Fixes #225

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -265,13 +265,8 @@ module Hanami
       #   end
       def content_type
         return @content_type unless @content_type.nil?
-
-        if accept_header?
-          type = content_type_from_accept_header
-          return type if type
-        end
-
-        default_response_type || default_content_type || DEFAULT_CONTENT_TYPE
+        @content_type = content_type_from_accept_header if accept_header?
+        @content_type || default_response_type || default_content_type || DEFAULT_CONTENT_TYPE
       end
 
       # Action charset setter, receives new charset value
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -569,15 +569,40 @@ module Hanami
         end.compact.max&.format
       end
 
+      # @since 1.0.1
+      # @api private
       class RequestMimeWeight
         include Comparable
 
-        attr_reader :quality, :index, :mime, :format, :priority
+        # @since 1.0.1
+        # @api private
+        attr_reader :quality
+
+        # @since 1.0.1
+        # @api private
+        attr_reader :index
+
+        # @since 1.0.1
+        # @api private
+        attr_reader :mime
+
+        # @since 1.0.1
+        # @api private
+        attr_reader :format
+
+        # @since 1.0.1
+        # @api private
+        attr_reader :priority
+
+        # @since 1.0.1
+        # @api private
         def initialize(mime, quality, index, format = mime)
           @quality, @index, @format = quality, index, format
           calculate_priority(mime)
         end
 
+        # @since 1.0.1
+        # @api private
         def <=>(other)
           return priority <=> other.priority unless priority == other.priority
           other.index <=> index
@@ -585,6 +610,8 @@ module Hanami
 
         private
 
+        # @since 1.0.1
+        # @api private
         def calculate_priority(mime)
           @priority ||= (mime.split('/'.freeze, 2).count('*'.freeze) * -10) + quality
         end
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -560,33 +560,38 @@ module Hanami
       # @see https://github.com/rack/rack/pull/659
       # @see https://github.com/hanami/controller/issues/59
       # @see https://github.com/hanami/controller/issues/104
+      # @see https://github.com/hanami/controller/issues/275
       def best_q_match(q_value_header, available_mimes)
-        values = ::Rack::Utils.q_values(q_value_header)
-        values = values.map do |req_mime, quality|
-          if req_mime == DEFAULT_ACCEPT
-            # See https://github.com/hanami/controller/issues/167
-            match = default_content_type
-          else
-            match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
-          end
+        values = []
+        ::Rack::Utils.q_values(q_value_header).each_with_index do |(req_mime, quality), index|
+          match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
           next unless match
-          [match, quality]
-        end.compact
-
-        if Hanami::Utils.jruby?
-          # See https://github.com/hanami/controller/issues/59
-          # See https://github.com/hanami/controller/issues/104
-          values.reverse!
-        else
-          # See https://github.com/jruby/jruby/issues/3004
-          values.sort!
+          values << Specification.new(req_mime, quality, index, match)
         end
 
-        value = values.sort_by do |match, quality|
-          (match.split('/'.freeze, 2).count('*'.freeze) * -10) + quality
-        end.last
+        value = values.sort.last
 
-        value.first if value
+        if value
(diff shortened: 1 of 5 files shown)


mereghost:
Some benchs:

Memory Allocation between `Rack` and `Action::Mime` `best_q_match`, when called with `'application/json,text/html,audio/*;q=0.8,text/css;q=0.3,*/*'` and `Hanami::Mimes::MIME_TYPES.values`
```
Comparison:
              Hanami:      17544 allocated
                Rack:      18480 allocated - 1.05x more
```

And some speed:
```
Comparison:
                Rack:    18312.6 i/s
              Hanami:    17634.0 i/s - same-ish: difference falls within error
```

lib/hanami/action/mime.rb
```diff
@@ -565,33 +560,41 @@ def content_type_with_charset
...
 
-        value.first if value
+        attr_reader :quality, :index, :mime, :format, :priority
+        def initialize(mime, quality, index, format = mime)
+          @quality, @index, @format = quality, index, format
+          calculate_priority(mime)
+        end
+
+        def <=>(other)
+          return priority <=> other.priority unless priority == other.priority
```
TiteiKo: I'm guessing you don't compare quality here because it's already taken into account in the priority formula?

lib/hanami/action/mime.rb
```diff
@@ -565,33 +560,41 @@ def content_type_with_charset
...
         end
 
-        value = values.sort_by do |match, quality|
-          (match.split('/'.freeze, 2).count('*'.freeze) * -10) + quality
-        end.last
+        value = values.max
+        value.format if value
+      end
+
+      class Specification
```
TiteiKo: `RequestMimeWeight`?

lib/hanami/action/mime.rb
```diff
@@ -565,33 +560,41 @@ def content_type_with_charset
...
-        values = ::Rack::Utils.q_values(q_value_header)
-        values = values.map do |req_mime, quality|
-          if req_mime == DEFAULT_ACCEPT
-            # See https://github.com/hanami/controller/issues/167
-            match = default_content_type
-          else
-            match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
-          end
+        values = []
+        ::Rack::Utils.q_values(q_value_header).each_with_index do |(req_mime, quality), index|
```
TiteiKo: You could make your code a bit more elegant:

```ruby
     def best_q_match(q_value_header, available_mimes)
        values = ::Rack::Utils.q_values(q_value_header).each_with_index.map do |(req_mime, quality), index|
          match = available_mimes.find { |am| ::Rack::Mime.match?(am, req_mime) }
          Specification.new(req_mime, quality, index, match) if match
        end

        values.compact.max&.format
      end
```

mereghost: Gah! I keep forgetting I can chain `Enumerator` calls. Shame on me for still having some ruby1.8 reflexes. 

That would leave the atribution to values a moot point and I could chain the `compact.max&.format` to the end of the block. WDYT?

lib/hanami/action/mime.rb
mereghost: I'm really bad at naming things, so naturally your suggestion seems better. :D

lib/hanami/action/mime.rb
mereghost: Yep.

Since priority already factors the quality of the mime type, there's no need to compare based into it. TBH I starting by comparing it, ran some tests, got weird results and then realized that.

Oh, and thanks for the review @TiteiKo =)

lib/hanami/action/mime.rb
TiteiKo: Yes `values` is useless, I've let it on another line for readability but hesitated, so if you find it understandable enough, just chain to the end :+1:

lib/hanami/action/mime.rb
```diff
@@ -565,33 +560,41 @@ def content_type_with_charset
...
+        attr_reader :quality, :index, :mime, :format, :priority
+        def initialize(mime, quality, index, format = mime)
+          @quality, @index, @format = quality, index, format
+          calculate_priority(mime)
+        end
+
+        def <=>(other)
+          return priority <=> other.priority unless priority == other.priority
+          other.index <=> index
+          # return other.quality <=> quality unless quality == other.quality
```
TiteiKo: :ghost:

lib/hanami/action/mime.rb
```diff
@@ -565,33 +560,37 @@ def content_type_with_charset
...
+          calculate_priority(mime)
         end
 
-        value = values.sort_by do |match, quality|
-          (match.split('/'.freeze, 2).count('*'.freeze) * -10) + quality
-        end.last
+        def <=>(other)
+          return priority <=> other.priority unless priority == other.priority
+          other.index <=> index
+          # return other.quality <=> quality unless quality == other.quality
```
TiteiKo: Please remove this commented code :ghost:

mereghost: /facepalm

Issue #22, opened 2014-06-25
Params validations


jodosha:
Right now, the `params` passed to `#call` are an instance of `Lotus::Action::Params` which is a nice wrapper around a Ruby Hash.

I would love to pass subclasses of `Params` which are aware of the specific use case and they should be able to do:
- [x] White label filtering: only the declared params will be available in the final object passed to `#call`.
- [ ] Coercions (optional feature): coerce a single param to the given Ruby class.
- [ ] Validations: if the given params aren't valid, halt with an HTTP status.

This approach has the following advantages:
- Avoid the need for low level entities to have mass assignment protection features.
- Save CPU cycles in case of validation failures: there is no need to hit lower layers if the data isn't valid.

We probably need to add this interface:

``` ruby
class Lotus::Action::Params
  def self.param(name, type, options = {})
    # ...
  end

  def initialize(params)
    @params = params
  end

  def validate!
    halt 412 unless valid?
  end

  def valid?
    # ...
  end
end
```

And implement one of those solutions:
### A

``` ruby
class Signup
  include Lotus::Action
  params SignupParams

  def call(params)
    puts params.class # => SignupParams
  end
end

class SignupParams < Lotus::Action::Params
  param :first_name, String, presence: true
  param :last_name,  String, presence: true
  param :email,      String, presence: true, format: /(.*)/
end
```

Where:

``` ruby
def self.params(params_class)
  before do |params|
    params_class.new(params).validate!
  end
end
```
### B

``` ruby
class Signup
  include Lotus::Action

  params do
    param :first_name, String, presence: true
    param :last_name,  String, presence: true
    param :email,      String, presence: true, format: /(.*)/
  end

  def call(params)
    puts params.class # => Signup::Params
  end
end
```

Where `.params` would generate an inner class of the action `Signup::Params`, which inherits from `Lotus::Action::Params`.

sarahhenkens:
@jodosha The article is great on explaining what might go wrong. But what about a solution? Imagine a user model with a `username` and `email` field, both of these need to be unique. 

I do a POST and submit duplicate content for both fields. Relying on constraint exceptions will only trigger the first found constraint violation. Resulting in only it invalidating the first field. The user has to submit it again with the modified username to get the second validation constraint. 

Why not both? Add a uniqueness check in the validations block and also handle constraint violations properly in case the race condition DOES happen.

beauby:
@sarahhenkens I believe you meant to comment on #232.

[...]

PR #232, opened 2017-08-30
Introduce Hanami::Action::Params::Errors#add

jodosha:
This is a proposal to introduce a way to accumulate errors that cannot be handled directly by params validations.

This is mainly demanded by the need of having **only one object** in the view/templates to check if an operation was successful and eventually display the error messages. Respectively `params.valid?` and `params.error_messages`.

```erb
<%=
  form_for ... do
    div do
      unless params.valid?
        h2 "Invalid data"

        ul do
          params.error_messages.each do |message|
            li message
          end
        end
      end
    
    # ...
  end
%>
```

---

But with this example above we can't display other errors, that aren't defined by `params` block and its validations.

One example is the lack of _uniqueness validation_ and how people wrongly worked around this constraint. I've seen stuff like this:

```ruby
params do
  required(:book).schema do
    required(:code) { |value| str? & BookRepository.new.not_exist_by_code?(value) }
  end
end
```

This _poor man_ implementation reintroduces the same race condition that we wanted to avoid by not implementing _uniqueness validation_.

---

At the same time we want to still offer **only one object** to check in the view/templates, and this still can be `params.errors`.

Here's a full example:

```ruby
module Web::Controllers::Books
  class Update
    include Web::Action

    params do
      required(:book).schema do
        required(:code).filled(:str?)
      end
    end

    def call(params)
      return unless params.valid?

      book = BookRepository.new.update_by_author(params[:author_id], params[:id], params[:book])

      flash[:notice] = "Book was successfully updated"
      redirect_to routes.author_book_path(author_id: book.author_id, id: book.id)
    rescue Hanami::Model::UniqueConstraintViolationError
      params.errors.add(:book, :code, "is not unique")
    end
  end
end
```

By adding an error to `params.errors` we will get the ERB snippet above to work as expected and to show the message `"Code is not unique"`.

---

⚠️⚠️⚠️ 
I'm not thrilled by the idea of mutating `params.errors` and to use it as accumulator for all the errors that may happen during the execution of an action. Think of the ACL violations, or business rules, as the _uniqueness constraint_ they have nothing to do with HTTP params.

But this is an non-breaking change to solve this problem, @hanami/core if you have a better solution, please speak up.

For Hanami 2.0 I'm experimenting totally different flow to handle errors and exceptions.
⚠️ ⚠️ ⚠️


lib/hanami/action/params.rb
```diff
@@ -16,6 +16,62 @@ module Action
...
+        #       BookRepository.new.create(params[:book])
+        #     rescue Hanami::Model::UniqueConstraintViolationError
+        #       # 2. Add an error in case the record wasn't unique
+        #       params.errors.add(:book, :isbn, "is not unique")
+        #     end
+        #   end
+        def add(*args)
+          *keys, key, error = args
+
+          (if keys.empty?
```
davydovanton: WDYT, can we move this if condition to a private method? (for better readability)

lib/hanami/action/params.rb
```diff
@@ -129,7 +184,7 @@ def error_messages(error_set = errors)
       # @example
       #   params.valid? # => true
       def valid?
-        @result.success?
+        errors.empty?
```
davydovanton: WDYT about `@result.success? && errors.empty?`?

jodosha: @davydovanton I've considered it, but given when we initialize `Errors` it's already full of errors if `@result.success?` is `false`. Because if there is a validation error, `@result.success?` is `false` and it has error messages that are filled into `Errors`.

If you look at the following prospect:

```
true && true => true # validations ok, no errors added
true && false => false # validations ok, but errors were manually added
false && true => false # impossible because if validations are failed, then we'll have errors filled
false && false => false # validations failed, these errors where copied to `Errors`. There could be potential errors added manually
```

If you look at the right side of the operation (after `&&`), that value is always the same of the result: if it's `true`, then the result is `true`, if `false`, the result is `false`. For this reason, I think `errors.empty?` is enough. Does it makes sense?

lib/hanami/action/params.rb
jodosha: Done: https://github.com/hanami/controller/pull/232/commits/92d2334eb5368acd79c10df5009b152dec02f0e8

lib/hanami/action/params.rb
davydovanton: great. thanks for explanation! 💓

beauby:
The merged code is fine, but I disagree on the rationale.
> This poor man implementation reintroduces the same race condition that we wanted to avoid by not implementing uniqueness validation.

The race condition is not an issue per-se, as it will happen sufficiently rarely that recovering from it with a standard "something went wrong, please try again" is ok (at least for stuff such as usernames, emails, etc.).

Also, as mentioned by @sarahhenkens [here](https://github.com/hanami/controller/issues/22#issuecomment-326754396), relying on DB uniqueness constraint violations will 1. prevent from returning multiple non-uniqueness constraint violations at once, and 2. relies on exceptions for control-flow which is generally a bad idea.

AlfonsoUceda:
@beauby I disagree with you and @sarahhenkens, the only way to ensure uniqueness validation is in your database because you have an index ensuring that. Uniqueness validations in application level layer are a lie, of course you can implement it and everything will be fine until you have some concurrency scenario and a duplicate will be saved in your database, but IMO the best way to enforce uniqueness in database is using index and avoid the application layer.

In this case, you have to rely in an exception, I think it's the only way to control that because you are using an external system, in this case a database.

beauby:
@AlfonsoUceda To clarify, I'm not advocating not to use DB uniqueness constraints, but rather to use **both** uniqueness constraints and uniqueness validations.
Moreover, you haven't replied to the main concern: how do you accommodate a use-case such as a signup form with a `username` and an `email` field, each with a unique constraint – how do you signify the user that both constraints were violated?

AlfonsoUceda:
@beauby yeah, in those cases I agree with you, It will be easier for the user to have a better UX, I don't think so in that case it will be concurrency so I think we could support uniqueness validations to allow do that.

sarahhenkens:
There are 2 parts in this use case:

 1. Enforcement
 2. Validation

Enforcement shouldn't be used to validate data, it's meant as a protective layer on top of your data storage to ensure that all these race conditions are considered. And as @beauby said before, both methods should be used, not just one of them.

It's like a parking structure, the validation is the counter in front of the building on how many free spots there are available. The enforcement is the actual free spot. The validation is for good user experience, the enforcement is for good data consistency.

AlfonsoUceda:
Thanks @sarahhenkens very good explained. I agree.

Issue #240, opened 2017-10-31
`after` callbacks not executed when using `send_file`


nerdinand:
It seems that `after` callbacks are not executed when using `send_file` in the action. `before` callbacks are being executed. 

Here's an example:

```ruby
module Web::Controllers::Thing
  class Show
    include Web::Action

    after :clean_up

    params do
      required(:file_name).filled(:str?)
    end

    def call(params)
      file_name = params[:file_name]

      self.headers.merge!(
        {
          'Content-Type' => 'application/pdf',
          'Content-Security-Policy' => "object-src 'self'",
          'Content-Disposition' => "attachment; filename=\"#{file_name}\""
        }
      )

      send_file(file_path(file_name))
    end

    private

    def clean_up(params)
      file_name = params[:file_name]
      path = Hanami.root.join('public', file_path(file_name))
      FileUtils.rm path
    end

    def file_path(file_name)
      file_path = ::File.join('uploads', 'thing_pdfs', file_name).to_s
    end
  end
end
```

I'm trying to delete the file after it's been downloaded in `clean_up`. Each file should only be downloadable once.

When using `after :clean_up`, `clean_up` is never called. If I try `before :clean_up` however, it is called.

nerdinand:
Any takers for this one?

cllns:
@nerdinand It looks like this is because [`send_file`](https://github.com/hanami/controller/blob/master/lib/hanami/action/rack.rb#L378-L388) calls [`halt`](https://github.com/hanami/controller/blob/35e6951a4ee244aeedecce71b89b3462fb5a19cd/lib/hanami/action/throwable.rb#L80-L125), which seems to bypass the `after` callbacks.

Are you willing to write a failing spec for this? That's the first step to fixing this. :)

nerdinand:
@cllns that much I already gathered, investigating the code. I don't have much time these days, only just enough to report bugs. But maybe I'll get around to it...

nerdinand:
Alright, #242 reproduces this. Not sure the spec is in the right place, but it proves the problem...

jodosha:
@nerdinand Would you mind to try #243 and see if it fixes the problem for you? Thanks.

PR #242, opened 2017-11-13
Failing spec for after callback after send_file

nerdinand:
Specs to reproduce #240:

* with after callback fails
* with before callback passes


jodosha:
@nerdinand Thanks for reproducing the bug. Closing this in favor of #243

PR #245, opened 2017-12-01
Remove global state

jodosha:
This PR is a rework of `hanami-controller`.

It brings immutability to actions and removes the global state. The hackish technique of "framework duplication" is removed. Everything is simpler and even less magic ✨ .

---

The immutability principle brings the advantage of instantiating the action once and reuse the same instance for each incoming request.

#### Hanami::Router integration

```ruby
# config.ru
require "hanami/router"
require "hanami/controller"

module Web
  module Controllers
    module Books
      class Show < Hanami::Action
        def call(*)
        end
      end
    end
  end
end

configuration = Hanami::Controller::Configuration.new
router = Hanami::Router.new(configuration: configuration, namespace: Web::Controllers) do
  get "/books/:id", "books#show"
end

run router
```

#### Basic usage

```ruby
class Show < Hanami::Action
  def call(req, res)
    res[:article] = ArticleRepository.new.find(req.params[:id])
  end
end

action = Show.new(configuration: configuration)
response = action.call(id: 23)

response[:article].class # => Article
response[:article].id # => 23

response.exposures.keys # => [:params, :article]
```

`Hanami::Action` has become a superclass to inherit from. This will enforce Single Responsibility Principle.

Because actions are immutable, it's no longer possible to assign instance variables (eg. `@article = ...`) or use `.expose` DSL. To _expose_ any object you can use `Hanami::Action::Response#[]=` and `#[]`.

The returning value of `Hanami::Action#call` is no longer a serialized Rack response, but an instance of `Hanami::Action::Response`.

#### Dependency Injection

```ruby
class Show < Hanami::Action
  def initialize(configuration:, repository: ArticleRepository.new)
    @repository = repository
    super(configuration: configuration)
  end

  def call(req, res)
    res[:article] = repository.find(req.params[:id])
  end

  private

  attr_reader :repository
end

configuration = Hanami::Controller::Configuration.new
action = Show.new(configuration: configuration, repository: ArticleRepository.new)
action.call(id: 23)
```

#### Callbacks

```ruby
class Show < Hanami::Action
  before :authenticate, :set_article

  def call(*)
  end

  private

  def authenticate
    # ...
  end

  # `req` and `res` in the method signature is optional
  def set_article(req, res)
    res[:article] = ArticleRepository.new.find(req.params[:id])
  end
end
```

or

```ruby
class Show < Hanami::Action
  before { ... } # do some authentication stuff
  before { |req, res| res[:article] = ArticleRepository.new.find(req.params[:id]) }

  def call(*)
  end
end
```

#### Exception handling

```ruby
class Show < Hanami::Action
  handle_exception RecordNotFound => 404

  def call(*)
    raise RecordNotFound
  end
end

action = Show.new(configuration: configuration)
action.call({}) # => [404, {}, ["Not Found"]]
```

You can also define custom handlers for exceptions.

```ruby
class Create < Hanami::Action
  handle_exception ArgumentError => :my_custom_handler

  def call(*)
    raise ArgumentError.new("Invalid arguments")
  end

  private

  def my_custom_handler(req, res, exception)
    res.status = 400
    res.body   = exception.message
  end
end

action = Create.new(configuration: configuration)
action.call({}) # => [400, {}, ["Invalid arguments"]]
```

Exception policies can be defined globally via configuration:

```ruby
configuration = Hanami::Controller::Configuration.new do |config|
  config.handle_exception RecordNotFound => 404
end

class Show < Hanami::Action
  def call(*)
    raise RecordNotFound
  end
end

action = Show.new(configuration: configuration)
action.call({}) # => [404, {}, ["Not Found"]]
```

Exception handling can be toggled with via a setting in configuration.
Single actions can specify a different behavior via method override.

```ruby
configuration = Hanami::Controller::Configuration.new do |config|
  config.handle_exceptions = true
end

# or

module Articles
  class Show < Hanami::Action
    def call(*)
      raise RecordNotFound
    end

    private

    def handle_exceptions?
      false
    end
  end
end

action = Articles::Show.new(configuration: configuration)
action.call({}) # => raises RecordNotFound
```

#### Cookies

You can read the original cookies sent from the HTTP client via `req.cookies`.
If you want to send cookies in the response, use `res.cookies`.

They are read as a Hash from Rack env:

```ruby
require "hanami/controller"
require "hanami/action/cookies"

class ReadCookiesFromRackEnv < Hanami::Action
  include Hanami::Action::Cookies

  def call(req, *)
    # ...
    req.cookies[:foo] # => "bar"
  end
end

action = ReadCookiesFromRackEnv.new(configuration: configuration)
action.call({"HTTP_COOKIE" => "foo=bar"})
```

They are set like a Hash:

```ruby
require "hanami/controller"
require "hanami/action/cookies"

class SetCookies < Hanami::Action
  include Hanami::Action::Cookies

  def call(*, res)
    # ...
    res.cookies[:foo] = "bar"
  end
end

action = SetCookies.new(configuration: configuration)
action.call({}) # => [200, {"Set-Cookie" => "foo=bar"}, "..."]
```

#### Sessions

Similarly to cookies, you can read the session sent by the HTTP client via
`req.session`, and also manipulate it via `res.ression`.

```ruby
require "hanami/controller"
require "hanami/action/session"

class ReadSessionFromRackEnv < Hanami::Action

  def call(req, *)
    # ...
    req.session[:age] # => "35"
  end
end

action = ReadSessionFromRackEnv.new(configuration: configuration)
action.call({ "rack.session" => { "age" => "35" } })
```

Values can be set like a Hash:

```ruby
require "hanami/controller"
require "hanami/action/session"

class SetSession < Hanami::Action

  def call(*, res)
    # ...
    res.session[:age] = 31
  end
end

action = SetSession.new(configuration: configuration)
action.call({}) # => [200, {"Set-Cookie"=>"rack.session=..."}, "..."]
```

#### Redirect

```ruby
class Create < Hanami::Action
  def call(*, res)
    # ...
    res.redirect_to "http://example.com/articles/23"
  end
end

action = Create.new(configuration: configuration)
action.call({ article: { title: "Hello" }}) # => [302, {"Location" => "/articles/23"}, ""]
```

#### Added

- `Hanami::Action::Request#session` to access the HTTP session as it was originally sent
- `Hanami::Action::Request#cookies` to access the HTTP cookies as they were originally sent

#### Changed

- `Hanami::Action` is a superclass
- `Hanami::Action#initialize` requires a `configuration:` keyword argument
- `Hanami::Action#initialize` returns a frozen action instance
- `Hanami::Action#call` accepts `Hanami::Action::Request` and `Hanami::Action::Response`
- `Hanami::Action#call` returns `Hanami::Action::Response`
- Removed `Hanami::Controller.configure`, `.configuration`, `.duplicate`, and `.load!`
- Removed `Hanami::Action.use` to mount Rack middleware at the action level
- `Hanami::Controller::Configuration` changed syntax from DSL style to setters (eg. `Hanami::Controller::Configuration.new { |c| c.default_request_format = :html }`)
- `Hanami::Controller::Configuration#initialize` returns a frozen configuration instance
- Removed `Hanami::Controller::Configuration#prepare`
- Removed `Hanami::Action.configuration`
- Removed `Hanami::Action.configuration.handle_exceptions` in favor of `#handle_exceptions?`
- Removed `Hanami::Action.configuration.default_request_format` in favor of `#default_request_format`
- Removed `Hanami::Action.configuration.default_charset` in favor of `#default_charset`
- Removed `Hanami::Action.configuration.format` to register a MIME Type for a single action. Please use the configuration.
- Removed `Hanami::Action.expose` in favor of `Hanami::Action::Response#[]=` and `#[]`
- Removed `Hanami::Action#status=` in favor of `Hanami::Action::Response#status=`
- Removed `Hanami::Action#body=` in favor of `Hanami::Action::Response#body=`
- Removed `Hanami::Action#headers` in favor of `Hanami::Action::Response#headers`
- Removed `Hanami::Action#accept?` in favor of `Hanami::Action::Request#accept?`
- Removed `Hanami::Action#format` in favor of `Hanami::Action::Response#format`
- Introduced `Hanami::Action#format` as factory to assign response format: `res.format = format(:json)` or `res.format = format("application/json")`
- Removed `Hanami::Action#format=` in favor of `Hanami::Action::Response#format=`
- Removed `Hanami::Action#request_id` in favor of `Hanami::Action::Request#id`
- Removed `Hanami::Action#parsed_request_body` in favor of `Hanami::Action::Request#parsed_body`
- Removed `Hanami::Action#head?` in favor of `Hanami::Action::Request#head?`
- Removed `Hanami::Action#status` in favor of `Hanami::Action::Response#status=` and `#body=`
- Removed `Hanami::Action#session` in favor of `Hanami::Action::Response#session`
- Removed `Hanami::Action#cookies` in favor of `Hanami::Action::Response#cookies`
- Removed `Hanami::Action#flash` in favor of `Hanami::Action::Response#flash`
- Removed `Hanami::Action#redirect_to` in favor of `Hanami::Action::Response#redirect_to`
- Removed `Hanami::Action#cache_control`, `#expires`, and `#fresh` in favor of `Hanami::Action::Response#cache_control`, `#expires`, and `#fresh`, respectively
- Removed `Hanami::Action#send_file` and `#unsafe_send_file` in favor of `Hanami::Action::Response#send_file` and `#unsafe_send_file`, respectively
- Removed `Hanami::Action#errors`
- `Hanami::Action` callback hooks now accept `Hanami::Action::Request` and `Hanami::Action::Response` arguments
- `Hanami::Action` exception handlers now accept `Hanami::Action::Request`, `Hanami::Action::Response`, and exception arguments

---

Ref https://github.com/hanami/router/pull/158


lib/hanami/action/request.rb
```diff
@@ -1,44 +1,58 @@
...
 
-      # @raise [NotImplementedError]
-      #
-      # @since 0.3.1
-      # @api private
-      def session
-        raise NotImplementedError, 'Please include Action::Session and use Action#session'
+      def id
+        # FIXME make this number configurable and document the probabilities of clashes
+        @id ||= @env[REQUEST_ID] = SecureRandom.hex(DEFAULT_ID_LENGTH)
```
AlfonsoUceda: is not better a UUID by default?

README.md
```diff
@@ -892,53 +823,46 @@ Hanami::Controller is shipped with an extensive list of the most common MIME typ
...
 ```
 
 ### Streamed Responses
 
 When the work to be done by the server takes time, it may be a good idea to stream your response. Here's an example of a streamed CSV.
 
 ```ruby
-Hanami::Controller.configure do
-  format csv: 'text/csv'
-  middleware.use ::Rack::Chunked
```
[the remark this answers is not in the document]
jodosha: With this PR is no longer possible to use Rack middleware for single actions.

lib/hanami/action/request.rb
jodosha: It's a tradeoff, UUID has nearly no chances to have collisions, but it's slower. That's why we need a fast enough algorithm with reasonable low collision probabilities.

Also, I'd make it somehow configurable (maybe via a lambda), so devs can provide their own strategy if they have custom needs.

Does it makes sense?

AlfonsoUceda: yes, it's perfect ;) I just wanted to know the reason of use that :)

jodosha: For the record, this is the benchmark: https://gist.github.com/jodosha/a00192eb0b4108c3eab8b542b9f1ad39

UUID is almost 2.4x times slower than HEX 16.

lib/hanami/action.rb
```diff
@@ -28,44 +32,643 @@ module Hanami
...
+    def self.inherited(base)
       base.class_eval do
-        include Rack
-        include Mime
-        include Redirect
-        include Exposable
-        include Throwable
-        include Callbacks
+        include Utils::ClassAttribute
+        class_attribute :before_callbacks
```
mereghost: Quick question: callbacks are included for making easier to share behaviour?

jodosha: yes

mereghost: Ok, I can live with that. 

Callbacks always make me antsy. I'd rather call `super` (or some other collaborator) on `#call`

PR #247, opened 2017-12-27
Implement content_type guard

### LINKED ISSUE #236: REQ/RES Content-Type header clash
Hi guys,
First of all wanna say that project you've built is awesome, lightweight and clear.

### Issue

I've found that inside action I can't get easy access to the `@_env['CONTENT_TYPE']` like I can for `accept`. I start to dig and I've found 2 things.

1. [Hanami::Action::Request](https://github.com/hanami/controller/blob/master/lib/hanami/action/request.rb) is aimed to keep compatibility to the `Rack::Request`
2. At the same time `Hanami::Action::Mime#content_type` is presenting computed value of the response content type

n.1 is kind-of abused because of [Rack::Request#content_type](http://www.rubydoc.info/gems/rack/Rack/Request/Helpers#content_type-instance_method) presenting request content type.

n.2 is confusing because you expect it to be a request content type.

### Missing feature 

I found it because I was looking for `accept`-like feature, but for incoming payload. As you know `Content-Type` header for requests like `POST`, `PUT`, `PATCH` and so on are defining a content type of request payload and it would be nice to have 2-way contracts client-server, like

```ruby
class UserController::Update
  include Hanami::Action

  accept :json
  content_type :json

  def call(params); end
end
```

In case of `accept` we are checking that client accepts defined content-type of the response and with `content_type` we are checking that client will send to the server a payload of a specific content type.

### Contribution

I would like to help the project, but I not sure what we can do in this case. We have `Hanami::Action::Mime#format` to overwrite the response content type and we have `Hanami::Action::Mime#content_type` which is more like helper for the response. And if I will change it to request content type it will be a breaking change.

So if someone from the core-team can highlight an approach you will accept, I will try to open PR

Strech:
Closes #236 

Allows to restrict the `Content-Type` of the payload and halt with `415 Unsupported Media Type`

Example

```ruby
class Upload
  include Hanami::Action
  content_type :json

  def call(params)
    # ...
  end
end
```

Related latest RFC

1. https://tools.ietf.org/html/rfc7231#section-3.3
2. https://tools.ietf.org/html/rfc7231#section-4.3.1
3. https://tools.ietf.org/html/rfc7231#section-6.5.13

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -198,7 +198,7 @@ module Hanami
           before do
             mime_type = @_env[HTTP_CONTENT_TYPE] || default_content_type || DEFAULT_CONTENT_TYPE
 
-            unless mime_types.find {|mt| ::Rack::Mime.match?(mime_type, mt) }
+            if mime_types.none? {|mt| ::Rack::Mime.match?(mime_type, mt) }
               halt 415
             end
           end
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -18,7 +18,7 @@ module Hanami
 
       # The key that returns content mime type from the Rack env
       #
-      # @since 1.1.1
+      # @since x.x.x
       # @api private
       HTTP_CONTENT_TYPE    = 'CONTENT_TYPE'.freeze
 
@@ -176,7 +176,7 @@ module Hanami
         #
         # @param formats[Array<Symbol>] one or more symbols representing mime type(s)
         #
-        # @since 1.1.1
+        # @since x.x.x
         #
         # @example
         #   require 'hanami/controller'
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -16,6 +16,12 @@ module Hanami
       # @api private
       HTTP_ACCEPT          = 'HTTP_ACCEPT'.freeze
 
+      # The key that returns content mime type from the Rack env
+      #
+      # @since 1.1.1
+      # @api private
+      HTTP_CONTENT_TYPE    = 'CONTENT_TYPE'.freeze
+
       # The header key to set the mime type of the response
       #
       # @since 0.1.0
@@ -162,6 +168,38 @@ module Hanami
           before do
             unless mime_types.find {|mt| accept?(mt) }
               halt 406
+            end
+          end
+        end
+
+        # Restrict the payload type to the specified mime type symbols.
+        #
+        # @param formats[Array<Symbol>] one or more symbols representing mime type(s)
+        #
+        # @since 1.1.1
+        #
+        # @example
+        #   require 'hanami/controller'
+        #
+        #   class Upload
+        #     include Hanami::Action
+        #     content_type :json
+        #
+        #     def call(params)
+        #       # ...
+        #     end
+        #   end
+        #
+        #   # When called with "text/html"        => 415
+        #   # When called with "application/json" => 200
+        def content_type(*formats)
+          mime_types = formats.map { |format| format_to_mime_type(format) }
+
+          before do
+            mime_type = @_env[HTTP_CONTENT_TYPE] || default_content_type || DEFAULT_CONTENT_TYPE
+
+            unless mime_types.find {|mt| ::Rack::Mime.match?(mime_type, mt) }
+              halt 415
             end
           end
         end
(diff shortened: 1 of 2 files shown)


lib/hanami/action/mime.rb
```diff
@@ -16,6 +16,12 @@ module Mime
       # @api private
       HTTP_ACCEPT          = 'HTTP_ACCEPT'.freeze
 
+      # The key that returns content mime type from the Rack env
+      #
+      # @since 1.1.1
```
Strech: Not sure about that 🤔

lib/hanami/action/mime.rb
```diff
@@ -165,6 +171,38 @@ def accept(*formats)
             end
           end
         end
+
+        # Restrict the payload type to the specified mime type symbols.
+        #
+        # @param formats[Array<Symbol>] one or more symbols representing mime type(s)
+        #
+        # @since 1.1.1
```
Strech: Not sure about that 🤔

lib/hanami/action/mime.rb
cllns: You can write `x.x.x` instead, and the release process will replace it with the correct version :)

Strech:
@jodosha Good day, just to remind you about this PR, once you have a time, please give me some feedback ❤️

Strech:
Hola, any update?

lib/hanami/action/mime.rb
```diff
@@ -165,6 +171,38 @@ def accept(*formats)
...
+        #
+        #   # When called with "text/html"        => 415
+        #   # When called with "application/json" => 200
+        def content_type(*formats)
+          mime_types = formats.map { |format| format_to_mime_type(format) }
+
+          before do
+            mime_type = @_env[HTTP_CONTENT_TYPE] || default_content_type || DEFAULT_CONTENT_TYPE
+
+            unless mime_types.find {|mt| ::Rack::Mime.match?(mime_type, mt) }
```
cllns: Is this the same as:
```ruby
if mime_types.none? {|mt| ::Rack::Mime.match?(mime_type, mt) }
```
We're not using the resulting value, so I think a boolean reveals the intent better. Also, using `none?` instead of `any?` lets us use an `if` instead of `unless` which is :100: for me :)

I'm not 100% sure this is the same logic but I think it's equivalent?

Strech: Thanks, will add a test and check can we replace it with `none` or so 👍

Strech:
@cllns Many thanks for a good catch. Specs and a guard clause is adjusted

Strech:
@AlfonsoUceda Thanks 👍 Looking forward to helping you guys more

Strech:
@cllns @jodosha Hey guys! Do you have time to leave your feedback about that functionality? Thanks ❤️

AlfonsoUceda:
@Strech I'll test as soon as possible, I hope today.

AlfonsoUceda:
I've tested in a fresh app and it works great ;)

Strech:
@AlfonsoUceda Great, happy to hear 👏

AlfonsoUceda:
@jodosha I wouldn't wait until 2.0, I think this feature is good to merge it now, IMO we can add a DEPRECATION WARNING for `accept` but explaining that the implementation will be `content_type`'s implementation.
1.2 -> `accept` and `content_type` live together but with the deprecation warning in `accept`
1.3 -> remove `content_type` and move its implementation to `accept` but creating an alias.
1.4 -> remove the alias so we couldn't access to `content_type`

I don't know if it makes sense.

Strech:
@jodosha Yes, finally you see my point out of here https://github.com/hanami/controller/issues/236

Technically, the logical one is `accept`, it should guard us against taking something we don't want to accept. I was calling it "clash", but it's exactly what you've written.

At the same time, we can make a smooth migration from `content_type` to `accept` as @AlfonsoUceda described  👍

mereghost:
@jodosha Hmmm while I do agree that developers might get confused on which method use, `Content-Type` and `Accept` are very different beasts. The former is used to tell the server what the heck is the payload of the *request*, the latter is used to tell the server the format in which you want the *response*.  

`Accept` is specially useful on APIs that can return XML, JSON and whatever other format without having a weird file extension on your URL. We could alias accept to something as `responds_with` so the user get it's about the response.

`Content-Type` is useful as a body parsing/validation thing, so you are sure that you aren't receiving gibberish (from your application PoV). This one could become `accepts` since it has to deal with the request.

Quick note: The error code for a non-compliant accept header is 406. The error code for a non compliant content type is 415.

jodosha:
Thank you all for the precious feedback. 💚 
Let's ship this for now, then for 2.0 we can replace `.accept` implementation with `.content_type` implementation and remove the latter. Or, we can rename `.content_type` into `accepted_formats :json`, or `formats :json` to make it clear that it hasn't nothing to do with `Accept`.

PR #248, opened 2018-01-10
Fix conditions to add Last-Modified / Etag headers

unhappychoice:
http://hanamirb.org/guides/actions/http-caching/#etag

The document says if `If-None-Match` or `If-Modified-Since` headers are missing, `Etag` or `Last-Modified` headers will be in response headers.
And it is correct as common HTTP server specification, I think.

So, I fixed that.

### DIFF
--- a/lib/hanami/action/cache/conditional_get.rb
+++ b/lib/hanami/action/cache/conditional_get.rb
@@ -41,7 +41,7 @@ module Hanami
         # @since 0.3.0
         # @api private
         def header
-          { ETAG => @value }
+          { ETAG => @value } if @value
         end
 
         private
@@ -73,7 +73,7 @@ module Hanami
         # @since 0.3.0
         # @api private
         def header
-          { LAST_MODIFIED => @value.httpdate } if @value.respond_to?(:httpdate)
+          { LAST_MODIFIED => @value.httpdate } if @value && @value.respond_to?(:httpdate)
         end
 
         private
--- a/lib/hanami/action/cache/conditional_get.rb
+++ b/lib/hanami/action/cache/conditional_get.rb
@@ -41,7 +41,7 @@ module Hanami
         # @since 0.3.0
         # @api private
         def header
-          { ETAG => @value } if none_match
+          { ETAG => @value }
         end
 
         private
@@ -73,7 +73,7 @@ module Hanami
         # @since 0.3.0
         # @api private
         def header
-          { LAST_MODIFIED => @value.httpdate } if modified_since
+          { LAST_MODIFIED => @value.httpdate } if @value.respond_to?(:httpdate)
         end
 
         private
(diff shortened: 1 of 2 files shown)


lib/hanami/action/cache/conditional_get.rb
```diff
@@ -41,7 +41,7 @@ def fresh?
         # @since 0.3.0
         # @api private
         def header
-          { ETAG => @value } if none_match
```
AlfonsoUceda: Please, can you add `if @value`? if we send a ETAG value without value there will be a rack lint error because ETAG will be nil. IMO we should send this ETAG if we are using `fresh` method with `:etag` option.

lib/hanami/action/cache/conditional_get.rb
```diff
@@ -73,7 +73,7 @@ def fresh?
         # @since 0.3.0
         # @api private
         def header
-          { LAST_MODIFIED => @value.httpdate } if modified_since
+          { LAST_MODIFIED => @value.httpdate } if @value.respond_to?(:httpdate)
```
AlfonsoUceda: Please can you add `if @value && @value.respond_to?(:httpdate)`? Thanks

unhappychoice:
Thanks for reviewing :smile:
I will make commit for that soon.

unhappychoice:
Sorry for being late, I fixed the things.
So, would you like to review again please ? @AlfonsoUceda

AlfonsoUceda:
@unhappychoice thanks so much, I'm going to review as soon as possible but it looks good ;)

spec/integration/hanami/controller/cache_spec.rb
```diff
@@ -160,6 +161,15 @@ def call(_params)
       fresh etag: 'updated', last_modified: Time.now
     end
   end
+
+  class NilValue
+    include Hanami::Action
+    include Hanami::Action::Cache
+
+    def call(_params)
+      fresh etag: nil, last_modified: nil
```
AlfonsoUceda: Can you separate this test? I prefer in one action use `etag: nil` and in other action `last_modified: nil`.

AlfonsoUceda:
One last thing @unhappychoice, after the changes, can you rebase your `develop` branch and change the merge branch to master? Thanks!

jodosha:
Hi @unhappychoice, thanks for your PR. Can you please follow up, so we can have a better HTTP caching? Thank you very much in advance.

AlfonsoUceda:
@unhappychoice thanks for your work, I'm going to merge the changes and implement my request :)

AlfonsoUceda:
Completed with this commit: https://github.com/hanami/controller/commit/014236453ae1086b36b46d58cae28874fc57bd57

unhappychoice:
So sorry that I couldn't get completed and I appreciate your help :smile:

AlfonsoUceda:
No problem ;)

PR #250, opened 2018-02-19
Better missing session error

wheresmyjetpack:
This PR handles hanami/hanami#859.

When either `#session` or `#flash` are invoked by an action that does not include `Hanami::Action::Session`, raise `Hanami::Controller::MissingSessionError` to provide better insight into why those methods are undefined.

The docs can be updated (as mentioned in the above issue) to point out the need to include the session module, and the appropriate url can be placed in the new exception's message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ sudo: false
 cache: bundler
 script: 'script/ci'
 rvm:
-  - 2.3.5
+  - 2.3.6
   - 2.4.2
+  - 2.5.0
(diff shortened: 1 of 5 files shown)


AlfonsoUceda:
Thanks @wheresmyjetpack, I'll test tomorrow if I can. I'm going to keep `develop` branch becauase there is just one commit because it seems this branch comes from master.

wheresmyjetpack:
@AlfonsoUceda :+1: thanks

PR #262, opened 2018-04-25
deprecate parsed_request_body

### LINKED ISSUE #254: Deprecate `Hanami::Action#parsed_request_body`
Body parsers will be [removed](https://github.com/hanami/router/issues/163) from the internal of the framework in favor of a [Rack middleware](https://github.com/hanami/router/issues/162).

There isn't should be anymore distinctions between params and parsed body, as both are already merged in `Action#params`.

  - [ ] Use `Hanami::Utils::Deprecation`

---

This change **MUST** go to `develop` branch.

---

Ref https://github.com/hanami/router/issues/162

GustavoCaso:
Fixes #254 

@jodosha one question usually when deprecating a method we show a deprecation message and call the new method but since the new method actually is an argument of the method `call`from the action I wasn't sure how to proceed, we could call `params` that will return the whole Rack env or even constructor the params for the user our self, but I do not know if it will have some penalties. WDYT?

### DIFF
--- a/spec/unit/hanami/action_spec.rb
+++ b/spec/unit/hanami/action_spec.rb
@@ -1,3 +1,5 @@
+require "hanami/devtools/unit"
+
 RSpec.describe Hanami::Action do
   describe ".configuration" do
     after do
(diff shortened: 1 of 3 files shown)


spec/unit/hanami/action_spec.rb
```diff
@@ -136,6 +136,16 @@ def call(_)
   end
 
   describe "#parsed_request_body" do
+    before do
+      @original_stderr = $stderr
+      $stderr = File.open(File::NULL, "w")
+    end
+
```
GustavoCaso: @jodosha what do you think about what @AlfonsoUceda mentioned in of moving this functionality to `Hanami::Utils::IO`?

jodosha: @GustavoCaso Please have a look at https://github.com/hanami/utils/blob/develop/spec/support/silence_deprecations.rb

I'd add it to `hanami-devtools`, WDYT?

PR #263, opened 2018-04-26
Remove body cleanup for head

### LINKED ISSUE #253: Remove body cleanup for HEAD HTTP requests
`Hanami::Action` 1.x empties the body if the HTTP request is using HEAD verb.

While this is formally correct, and makes the action "full featured", we have the same control in `hanami-router` and `hanami`. This makes a total of three times for each request. The idea is to use `Rack::Head` from `rack` gem instead of these custom controls.

Developers who are using `hanami-controller` as standalone gem, can mount that middleware.
Developers who are using `hanami` don't need to take any action, because we'll mount that middleware at the project level.

---

This change **MUST** go to `unstable` branch.

---

Ref https://github.com/hanami/router/issues/164
Ref https://trello.com/c/fQw5F4Hp

GustavoCaso:
Fixes #253 

@jodosha I remove the cleanup of the `body` but looks like we still do some cleanup on the `headers` at the moment I left in here, but maybe we should consider moving it to a middleware or the router. WDYT?

### DIFF
--- a/spec/integration/hanami/controller/full_stack_spec.rb
+++ b/spec/integration/hanami/controller/full_stack_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe "Full stack application" do
   it "only allows entity headers if the request is HEAD" do
     head "/head", {}, "HTTP_ACCEPT" => "text/html"
 
-    expect(last_response.body).to_not    be_empty
+    expect(last_response.body).to_not be_empty
     expect(last_response.headers).to_not have_key("X-Renderable")
   end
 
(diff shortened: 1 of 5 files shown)


spec/integration/hanami/controller/full_stack_spec.rb
```diff
@@ -14,13 +14,6 @@ def app
     expect(last_response.body).to include(':greeting=>"Hello"')
   end
 
-  it "omits the body if the request is HEAD" do
```
jodosha: @GustavoCaso Instead of removing this spec, I'd keep it, but update with the new behavior.

GustavoCaso:
@jodosha I updated the code and the build is green 💚

Issue #266, opened 2018-06-11
Flash #empty? returns true when it's not (after a re-direct)


cllns:
Here's an example app illustrating the problem: https://github.com/cllns/flash-error-example-app

If you visit `/books`, you're redirected to `/home` and you'll see this rendered (from both the layout template and view template:

```
Value of flash[:info]: Books are gone! Go home!

Value of flash.empty?: true 
```
Empty should not be true if there are values in the flash.

I tried to figure this out for a couple hours and I'm stuck.

I added some debugging statements to `Hanami::Action::Flash#empty?` and it shows the following:

```
16:43:01 web.1  | #<Hanami::Action::Flash:0x7f9f299a0130 {:success=>"You have been signed out 👍"}>
16:43:01 web.1  | #empty? body definition
16:43:01 web.1  | _values: [{}]
16:43:01 web.1  | @merged: {:success=>"You have been signed out 👍"}
16:43:01 web.1  | Empty?: true
```
If I change the return value from `#empty?` to `_values.all?(&:empty?)) && @merged.empty?` then it works (since _values isn't being synced with @merged) in a way I'd expect it to. This does not break any specs, but I can't figure out the issue so that I could write a spec that fails without that change.

I'm not 100% sure this is a bug with `hanami-controller`, but I think it is. Specifically, I think it has something to do with the view being rendered after the controller has called `#finish`?

jodosha:
@cllns Can you please take care of this? Thanks.

GustavoCaso:
Fixed in https://github.com/hanami/controller/pull/267

PR #267, opened 2018-06-27
fix double flash bug, flash#empty?, flash#each and flash#map

### LINKED ISSUE #260: Regression: Return of the double-flash bug
It seems #196 has returned to haunt us.

Here's a minimal test app to reproduce this:
[flash_test.zip](https://github.com/hanami/controller/files/1937644/flash_test.zip)

1. Visit http://localhost:2300/. You will see:

2. Click on Edit. You will see a form:

3. Click on Save. As expected, the form will render again and show the success flash message:

4. Click on Cancel. The root page will render but show the flash message again!

Expected behaviour: The root page renders without showing the flash message.

As with #196, adding
```ruby
before { flash.clear }
```
to the index action is a workaround.

GustavoCaso:
Fixes #260 

@jodosha here is one implementation for fixing the double flash error.

Following our previous talk, I removed the code for `@merged` from #215 and added the logic to use `keep` us a way of telling flash to store the data for previous consumption if we trigger a redirect, also added a new spec in the integration folder to prove that the flash data is how we want it.

I see one issue with this approach, the case where we do not consume the key after the redirect, making it available for future access, we could also store the `request_id` and the `last_request_id` in the `kept_hash`

Looking forward to your comments.

### DIFF
--- a/lib/hanami/action/flash.rb
+++ b/lib/hanami/action/flash.rb
@@ -12,29 +12,26 @@ module Hanami
       # @api private
       SESSION_KEY = :__flash
 
-      # Session key where the last request_id is stored
+      # Session key where keep data is store for redirect
       #
-      # @since 0.4.0
+      # @since x.x.x
       # @api private
-      LAST_REQUEST_KEY = :__last_request_id
+      KEPT_KEY = :__kept_key
 
       # Initialize a new Flash instance
       #
       # @param session [Rack::Session::Abstract::SessionHash] the session
-      # @param request_id [String] the HTTP Request ID
       #
       # @return [Hanami::Action::Flash] the flash
       #
       # @see http://www.rubydoc.info/gems/rack/Rack/Session/Abstract/SessionHash
       # @see Hanami::Action::Rack#session_id
-      def initialize(session, request_id)
+      def initialize(session)
         @session         = session
-        @request_id      = request_id
-        @last_request_id = session[LAST_REQUEST_KEY]
-        @merged          = {}
+        @keep            = false
 
-        session[SESSION_KEY]             ||= {}
-        session[SESSION_KEY][request_id] ||= {}
+        session[KEPT_KEY] ||= []
+        session[SESSION_KEY] = {}
       end
 
       # Set the given value for the given key
@@ -45,7 +42,7 @@ module Hanami
       # @since 0.3.0
       # @api private
       def []=(key, value)
-        data[key] = value
+        _data[key] = value
       end
 
       # Get the value associated to the given key, if any
@@ -55,28 +52,26 @@ module Hanami
       # @since 0.3.0
       # @api private
       def [](key)
-        @merged.fetch(key) do
-          _values.find {|data| !data[key].nil? }
-        end
+        _data.fetch(key) { search_in_kept_data(key) }
       end
 
-      # Iterates through data
+      # Iterates through current request data and kept data
       #
       # @param blk [Proc]
       #
       # @since 1.2.0
       def each(&blk)
-        @merged.each(&blk)
+        _values.each(&blk)
       end
 
-      # Iterates through data
+      # Iterates through current request data and kept data
       #
       # @param blk [Proc]
       # @return [Array]
       #
       # @since 1.2.0
       def map(&blk)
-        @merged.map(&blk)
+        _values.map(&blk)
       end
 
       # Removes entirely the flash from the session if it has stale contents
@@ -91,11 +86,11 @@ module Hanami
         # this bug that I've found via a browser.
         #
         # It may happen that `#flash` is nil, and those two methods will fail
-        unless flash.nil?
-          expire_stale!
-          remove!
-          merge!
-          set_last_request_id!
+        unless _data.nil?
+          update_kept_request_count
+          keep_data if @keep
+          expire_kept
+          remove
         end
       end
 
@@ -107,48 +102,38 @@ module Hanami
       # @since 0.3.0
       # @api private
       def empty?
-        _values.all?(&:empty?)
+        _values.empty?
       end
 
       # @return [String]
       #
       # @since 1.0.0
       def inspect
-        "#<#{self.class}:#{'0x%x' % (__id__ << 1)} #{@merged.inspect}>"
+        "#<#{self.class}:#{'0x%x' % (__id__ << 1)} {:data=>#{_data.inspect}, :kept=>#{kept_data.inspect}} >"
+      end
+
+      # Set @keep to true, is use when triggering a redirect, and the content of _data is not empty.
+      # @return [TrueClass, NilClass]
+      #
+      # @since x.x.x
+      # @api private
+      #
+      # @see Hanami::Action::Flash#empty?
+      def keep!
+        return if empty?
+        @keep = true
       end
 
       private
 
-      # The flash registry that holds the data for **all** the recent requests
+      # The flash registry that holds the data for the current requests
       #
       # @return [Hash] the flash
       #
       # @since 0.3.0
       # @api private
-      def flash
+      def _data
         @session[SESSION_KEY] || {}
-      end
-
-      # The flash registry that holds the data **only for** the current request
-      #
-      # @return [Hash] the flash for the current request
-      #
-      # @since 0.3.0
-      # @api private
-      def data
-        flash[@request_id] || {}
-      end
-
-      # Expire the stale data from the previous request.
-      #
-      # @return [void]
-      #
-      # @since 0.3.0
-      # @api private
-      def expire_stale!
-        flash.each do |request_id, _|
-          flash.delete(request_id) if delete?(request_id)
-        end
       end
 
       # Remove the flash entirely from the session if empty.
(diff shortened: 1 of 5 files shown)


cllns:
Thanks for working on this! 

I just tried this branch on a local project (hoping this would also solve #266) but it looks like `#empty?` is returning `true` on the flash after a redirect, even when there are contents. Doing, for example `flash[:success]` works fine, but `flash.empty?` is returning true even when `flash[:success]` has a message. Also, even calling `#map` on the hash doesn't include the flash message from before the redirect. :/ 

Let me know if that's enough info or if you need more.

GustavoCaso:
@cllns Thank you so much for the comment.

I wasn't aware of #266 issue but I think we can tackle here as well. With the above comment, I have enough information to try fixing it here and add some specs.

GustavoCaso:
@cllns Do you mind having another try with your dummy app to see if the current code fixes #266 

@jodosha Also if you want to have a look, I'll be happy to hear both opinions about the code. I know the spec setup is not easy for testing flash and redirect but I hope we can tune in future iterations.

Looking forward to hearing your comments ❤️

GustavoCaso:
@jodosha @cllns I getting to a point where I'm quite happy with the current implementation. 

At the moment the code for flash management has been reduced quite a lot 😄 we no longer need to worry the `expire_stale_data` as well as `setting_last_request_id` that if you ask me, it made the code much more clear and more straightforward.

So how it works is by setting the flag `@keep` every time there is a `redirect_to` telling the flash to store the current `data` in our `KEPT_KEY` in the `session`. Once we try accessing the data with `[](key)` if the key is not present in the current flash data it will look for it in the `kept` hash, if the key is present in `kept` data it will return and delete at the same time this is our way to `expire` kept data.

At the moment I see one problem if a user sets the flash key and trigger a redirect and do not access that key, the key will be there until someone access it, making `flash#empty?` returns` false` and `flash#each` and `flash#map` to iterate over it as well. I did try adding some automatic expire like what we had with `last_request_id`, but before adding more logic, I wanted to know your thoughts on this.

jodosha:
Even if this is a bug fix, me and @GustavoCaso decided to have `develop` as target branch because it changes the internal implementation of `Flash`. The public API is unchanged, but still, it's worth to release this via the next beta release of August: 1.3.

jodosha:
@cllns Can you please manual test this in your app? Thanks.

cllns:
Just tried this out. Unfortunately, the flash messages aren't being expired _at all_ for me :( They're sticking around forever.

The reason is that I'm using `.each { |type, message| # ... }` to iterate through the messages, after a redirect. Since I'm not calling `#[:success]` directly on the flash, it's not expiring the `kept` message at all.

jodosha:
Temporary removing myself from the reviewers until we don't find a code solution. Thank you all.

cllns:
@GustavoCaso Just tried it again. The flash is now expiring properly, but I still have the original problem from #266 :/
`flash#empty?` is returning `true`
`flash#inspect` is returning `<Hanami::Action::Flash:0x7f940ff1ce58 {:success=>"You have been signed out 👍"}>` (not empty).

The original workaround that I had, calling `local(:flash).map(&:to_s).empty?` instead of `local(:flash).empty?` continues to work, but that's not a great solution since we should be able to just call `#empty?`.

cllns:
👍 Works great! Fixes #266 and the issue I reported last time :)

Great work!

lib/hanami/action/flash.rb
```diff
@@ -6,35 +6,59 @@ module Action
     # @since 0.3.0
     # @api private
     class Flash
+
+      class Kept
```
cllns: I think we should add comments to this class and the methods, as private api

lib/hanami/action/flash.rb
```diff
@@ -6,35 +6,59 @@ module Action
...
+      def initialize(session)
         @session         = session
-        @request_id      = request_id
-        @last_request_id = session[LAST_REQUEST_KEY]
-        @merged          = {}
+        @keep            = false
 
-        session[SESSION_KEY]             ||= {}
-        session[SESSION_KEY][request_id] ||= {}
+        session[KEPT_KEY]                ||= []
```
cllns: Some of these spaces and be removed. Might even be able to line these lines up with `@session` and `@keep`, if ya want

lib/hanami/action/flash.rb
```diff
@@ -92,10 +114,10 @@ def clear
         #
```
cllns: Do you have any idea about the `FIXME` on line 112? Might be a good time to tackle that if you can figure it out. It might not be relevant anymore with the refactoring that you did?

lib/hanami/action/flash.rb
```diff
@@ -107,112 +129,111 @@ def clear
...
+        new_kept = kept << Kept.new(flash)
+
+        @session[KEPT_KEY] = new_kept
       end
 
-      # Remove the flash entirely from the session if empty.
+      # Return value for key inside KEPT_KEY hash and delete it.
       #
-      # @return [void]
+      # @return [String|Integer|Nil] the current value for that key
```
cllns: These should be comma separated, and `Nil` -> `nil` :)

lib/hanami/action/flash.rb
```diff
@@ -107,112 +129,111 @@ def clear
...
+      def update_kept_request_count
+        kept.each(&:increase_request_count!)
       end
 
-      # Values from all the stored requests
+      def search_in_kept_data(key)
+        value_in_kept = kept.find { |kept| kept.key?(key) }
+        return unless value_in_kept
+
+        value_in_kept.find(key)
```
cllns: Can we change to:
```ruby
value_in_kept = kept.find { |kept| kept.key?(key) }
value_in_kept&.find(key)
```
Could even do one line, but I think that'd be confusing with double `find`.

lib/hanami/action/flash.rb
```diff
@@ -107,112 +129,111 @@ def clear
...
+      # The flash registry that holds the data for **all** the recent requests
       #
-      # @return [Array]
+      # @return [Hash] the flash
       #
       # @since 0.3.0
       # @api private
-      def _values
-        flash.select { |session_id, _| !delete?(session_id) }.values
+      def flash
```
cllns: I wonder if this should be `_data` or something? `Flash#flash` is a little strange?

spec/support/fixtures.rb
```diff
@@ -1520,8 +1520,8 @@ def initialize
         get '/head', to: 'home#head'
         resources :books, only: [:index, :create, :update]
 
-        get  '/settings', to: 'settings#index'
-        post '/settings', to: 'settings#create'
+        get  '/settings',     to: 'settings#index'
```
cllns: Can get rid of these spaces :)

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
-      # @return [Hash] the flash for the current request
+      # @return [void]
       #
       # @since 0.3.0
       # @api private
-      def data
-        flash[@request_id] || {}
+      #
+      # @see Hanami::Action::Flash#empty?
+      def remove!
```
mereghost: Any reason for this being a *bang* methods? I mean, in Ruby's stdlib bang methods are destructive methods and they only exist if a non-banged variant is even possible. Are there any better names that do not include a bang?

Not really a huge problem, but having different semantics across each piece of the framework + language, IMO, kinda break the POLS.

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
       #
-      # @return [Array]
+      # @return [Hash] the current value of KEPT_KEY
       #
-      # @since 0.3.0
+      # @since x.x.x
       # @api private
-      def _values
-        flash.select { |session_id, _| !delete?(session_id) }.values
+      def expire_kept!
```
mereghost: Same as above.

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
+      # @return [Hash] the current value of KEPT_KEY
       #
-      # @since 0.4.0
+      # @since x.x.x
       # @api private
-      #
-      # @see Hanami::Action::Flash#expire_stale!
-      def delete?(request_id)
-        ![@request_id, @session[LAST_REQUEST_KEY]].include?(request_id)
+      def update_kept_request_count!
```
mereghost: Same as above

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
-      # @return [void]
-      # @since 0.4.0
+      # @param new_kept_data
+      # @return [Hash] the current value of KEPT_KEY
+      #
+      # @since x.x.x
       # @api private
-      def set_last_request_id!
-        @session[LAST_REQUEST_KEY] = @request_id if !empty?
+      def update_kept!(new_kept_data)
```
mereghost: Same as above

lib/hanami/action/flash.rb
```diff
@@ -55,28 +52,26 @@ def []=(key, value)
       # @since 0.3.0
       # @api private
       def [](key)
-        @merged.fetch(key) do
-          _values.find {|data| !data[key].nil? }
-        end
+        _data[key] || search_in_kept_data(key)
```
jodosha: This should still use `fetch`.

What we're looking for here is the presence of `key` in the internal data. `fetch` yields the block if the key is missing, which is the wanted behavior.

---

The syntax with `||` on the other hand, yields the rightmost expression both if the key is missing (good), but also if the key is present but the value is `nil` (bad).

jodosha:
@mereghost @GustavoCaso I asked for the "bang methods" for consistency with the already existing methods. If you folks believe that it's the case to rename them, please do 👍

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
+      # @return [Hash] the current value of KEPT_KEY
       #
-      # @since 0.3.0
+      # @since x.x.x
       # @api private
-      def _values
-        flash.select { |session_id, _| !delete?(session_id) }.values
+      def expire_kept!
+        new_kept_data = kept.reject do |kept_data|
+          JSON.parse(kept_data)['count'] >= 2
```
jodosha: @GustavoCaso I think we should be defensive here and don't assume that this `count` is 1. present 2. an integer. If session data is modified (read evil craft), this would crash.

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
+      # @since x.x.x
       # @api private
-      #
-      # @see Hanami::Action::Flash#expire_stale!
-      def delete?(request_id)
-        ![@request_id, @session[LAST_REQUEST_KEY]].include?(request_id)
+      def update_kept_request_count!
+        new_kept_data = kept.map do |kept_data|
+          json_kept = JSON.parse(kept_data)
+          json_kept['count'] += 1
```
jodosha: Same of https://github.com/hanami/controller/pull/267/files#r200376689

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
+      #
+      # @return [Hash] the current value of KEPT_KEY
       #
-      # @see Hanami::Action::Flash#[]
-      def merge!
-        @merged = last_request_flash.merge(data)
+      # @since x.x.x
+      # @api private
+      def keep_data
+        new_kept_data = kept << { count: 0, data: _data }.to_json
```
jodosha: For consistency with https://github.com/hanami/controller/pull/267/files#diff-ce13f23d4ed57d06a91f48c61b42c054R209 this should be `JSON.generate` instead of using `to_json`.

lib/hanami/action/flash.rb
```diff
@@ -125,94 +133,124 @@ def inspect
...
       # @api private
-      #
-      # @see Hanami::Action::Flash#expire_stale!
-      def delete?(request_id)
-        ![@request_id, @session[LAST_REQUEST_KEY]].include?(request_id)
+      def update_kept_request_count!
+        new_kept_data = kept.map do |kept_data|
+          json_kept = JSON.parse(kept_data)
+          json_kept['count'] += 1
+          json_kept.to_json
```
jodosha: Same of https://github.com/hanami/controller/pull/267/files#r200377155

lib/hanami/action/rack.rb
```diff
@@ -213,16 +205,6 @@ def response
         [ @_status || DEFAULT_RESPONSE_CODE, headers, @_body || DEFAULT_RESPONSE_BODY.dup ]
       end
 
-      # Calculates an unique ID for the current request
-      #
-      # @return [String] The unique ID
-      #
-      # @since 0.3.0
-      def request_id
```
jodosha: ⚠️ This cannot be removed, it's public API and IIRC used by `hanami`.

lib/hanami/action/flash.rb
GustavoCaso: Great point @jodosha I will use `fetch`

lib/hanami/action/rack.rb
GustavoCaso: No problem I will restore it 😄

lib/hanami/action/flash.rb
GustavoCaso: @jodosha do you like the idea of:

```ruby
count = JSON.parse(kept_data)['count']
count.to_i >= 2
```

In the case `count` is not present using `to_i` will return `0` and ultimatley the condition will evaluate to `false`

lib/hanami/action/flash.rb
GustavoCaso: Unfortunately, I have no idea about it. Maybe @jodosha can give some information

GustavoCaso:
@jodosha @cllns @mereghost I have applied all feedback. Thank you so much for taking the time to review the PR. If we are happy with the end result we could squash into a single commit.

PR #268, opened 2018-07-03
Remove body cleanup for http status without body

### LINKED ISSUE #264: Remove body cleanup for HTTP responses that doesn't require a body (e.g. 204)
`Hanami::Action` 1.x empties the body if the HTTP response doesn't require a body (e.g. 204).

While this is formally correct, and makes the action "full featured", we have the same control in `hanami-router` and `hanami`. This makes a total of three times for each request. The idea is to give responsibility to developers to produce a compliant HTTP response, or to use a Rack middleware to ensure this correctness.

---

This change **MUST** go to `unstable` branch.

---

Ref https://github.com/hanami/controller/issues/253
Ref https://github.com/hanami/router/issues/164
Ref https://trello.com/c/fQw5F4Hp

GustavoCaso:
Fixes #264

### DIFF
--- a/spec/integration/hanami/controller/head_spec.rb
+++ b/spec/integration/hanami/controller/head_spec.rb
@@ -35,12 +35,36 @@ RSpec.describe "HTTP HEAD" do
 
   HTTP_TEST_STATUSES_WITHOUT_BODY.each do |code|
     describe "with: #{code}" do
-      it "does send body and default headers" do
-        get "/code/#{code}"
+      if [204, 304].include?(code)
+        it "doesn't send body and default headers" do
+          get "/code/#{code}"
 
-        expect(response.status).to           be(code)
-        expect(response.body).to_not         be_empty
-        expect(response.headers.to_a).to_not include(["X-Frame-Options", "DENY"])
+          expect(response.status).to           be(code)
+          expect(response.body).to             eq("")
+          expect(response.headers.to_a).to_not include(["X-Frame-Options", "DENY"])
+        end
+
+        it "doesn't send Content-Length header" do
+          get "/code/#{code}"
+
+          expect(response.status).to      be(code)
+          expect(response.headers).to_not have_key("Content-Length")
+        end
+      else
+        it "does send body and default headers" do
+          get "/code/#{code}"
+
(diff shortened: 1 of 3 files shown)


spec/support/fixtures.rb
```diff
@@ -9,7 +9,7 @@
 require 'hanami/action/glue'
 require_relative "./renderer"
 
-HTTP_TEST_STATUSES_WITHOUT_BODY = Set.new((100..199).to_a << 204 << 205 << 304).freeze
+HTTP_TEST_STATUSES_WITHOUT_BODY = Set.new((100..199).to_a << 205).freeze
```
GustavoCaso: @jodosha I had to change this `HTTP_STATUS` code because Rack by itself is cleaning the body for then https://github.com/rack/rack/blob/master/lib/rack/response.rb#L65 so the test was always failing. 

I haven't remove them from `HTTP_STATUSES_WITHOUT_BODY` because I wanted to hear your thoughts

spec/integration/hanami/controller/head_spec.rb
```diff
@@ -64,11 +64,11 @@ def response
         expect(response.headers["Content-Language"]).to eq("en")
       end
 
-      it "doesn't send Content-Length header" do
+      it "does send Content-Length header" do
         get "/code/#{code}"
```
GustavoCaso: Also, I did have to change this spec since we are no longer removing the body, the header is sent by `MockResponse`.

I was able to follow the code path:

1. Hanami does remove the header with `_empty_headers(res)`  https://github.com/hanami/controller/pull/268/files#diff-c3eb2b7cc639097f82ae3c2ed20be06bR714

2. `RackTest` at this line https://github.com/rack-test/rack-test/blob/master/lib/rack/mock_session.rb#L31
will return a `MockResponse` that had added the header. (My conclusion is the rack is adding the header again since there is a body)

WDYT? @jodosha

spec/support/fixtures.rb
jodosha: @GustavoCaso As discussed in our call, these HTTP statuses should stay.

PR #272, opened 2018-09-21
Fix: add missing 'json' require to action/flash

timriley:
If this file is required independently and 'json' isn't otherwise required already in the project, using flash will result in a "NameError: uninitialized constant JSON" error.

@jodosha ready for your review. Hopefully this is a quick one :)

Should this be targeting `develop` or `master`?


jodosha:
@timriley Thanks for this PR. I've restarted the build, because it was failing for `jruby-9.1` on CircleCI.

Fixes go to `master`.

PR #274, opened 2018-09-22
Remove module prepending from `Action.inherited` hook

timriley:
Opening this PR for discussion, since it achieves a few important things:

1. It makes it possible to create `Hanami::Action` subclasses that use dry-auto_inject to inject their own dependencies (the prepended `#initialize` was breaking this before)
2. It makes it possible to have deeper class hierarchies, such as `class ApplicationAction < Hanami::Action`, and then e.g. `class Users::Show < ApplicationAction`) (since we're no longer prepending `#initialize`, which in the previous implementation was broken when prepended more than once)
3. It improves understandability of the library - when a subclass implements `#handle`, but the main interface remains `#call`, then application developers will implicitly understand that some extra code is being run in between the two.

👋 @jodosha, I'll let you suggest who else to bring in for review here :)

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -179,7 +179,6 @@ module Hanami
         class_attribute :after_callbacks
         self.after_callbacks = Utils::Callbacks::Chain.new
 
-        prepend InstanceMethods
         include Validatable if defined?(Validatable)
       end
     end
@@ -425,39 +424,38 @@ module Hanami
       end
     end
 
-    # Callbacks API instance methods
+    # Implements the Rack/Hanami::Action protocol
     #
     # @since 0.1.0
     # @api private
-    module InstanceMethods
-      # Implements the Rack/Hanami::Action protocol
-      #
-      # @since 0.1.0
-      # @api private
-      def call(env)
-        request  = nil
-        response = nil
-        halted   = catch :halt do
-          begin
-            params   = self.class.params_class.new(env)
-            request  = Hanami::Action::Request.new(env, params)
-            response = Hanami::Action::Response.new(action: self.class.name, configuration: configuration, content_type: Mime.calculate_content_type_with_charset(configuration, request, accepted_mime_types), env: env, header: configuration.default_headers)
-            _run_before_callbacks(request, response)
-            super(request, response)
-            _run_after_callbacks(request, response)
-          rescue => exception
-            _handle_exception(request, response, exception)
-          end
-        end
+    def call(env)
+      request  = nil
+      response = nil
 
-        finish(request, response, halted)
+      halted = catch :halt do
+        begin
+          params   = self.class.params_class.new(env)
+          request  = Hanami::Action::Request.new(env, params)
+          response = Hanami::Action::Response.new(action: self.class.name, configuration: configuration, content_type: Mime.calculate_content_type_with_charset(configuration, request, accepted_mime_types), env: env, header: configuration.default_headers)
+          _run_before_callbacks(request, response)
+          handle(request, response)
+          _run_after_callbacks(request, response)
+        rescue => exception
+          _handle_exception(request, response, exception)
+        end
       end
+
+      finish(request, response, halted)
     end
 
     def initialize(**)
     end
 
     protected
+
+    def handle(request, response)
+      raise NotImplementedError, "Hanami::Action subclasses must implement #handle(request, response)"
+    end
 
     # Halt the action execution with the given HTTP status code and message.
     #
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -406,25 +406,30 @@ module Hanami
       handled_exceptions.merge!(exception)
     end
 
+    def self.new(configuration:, **args)
+      allocate.tap do |obj|
+        obj.instance_variable_set(:@configuration, configuration)
+        obj.instance_variable_set(:@accepted_mime_types, Mime.restrict_mime_types(configuration, accepted_formats))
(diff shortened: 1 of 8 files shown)


jodosha:
@timriley Thanks for this PR.

As per our chat discussion, I remember that prepending `InstanceMethods` is problematic for DI tools like `dry-auto_inject`. We must fix this problem.

---

My question here is about `#call`. Until now in `unstable`, that method is prepended by `InstanceMethods` as well. Does this causes problems to your integration? Can we still keep `#call` prepended?

timriley:
@jodosha Good question. I should have explained myself more.

The genesis idea came from [a mastodon discussion about unexpected behaviour being applied when using hanami-controller](https://ruby.social/@judofyr/100763452994365386). In response I mused a little on when best to prepend/not to prepend behaviour, and then Tom Dalling suggested what I felt was a much cleaner approach (to avoid prepending `#call`):

> If it were me, I'd use two different method names. Probably keep the external interface as `call`, and rename the internal method to `run` or `execute`. Then you don't need prepend, and it's a lot less surprising.

This really resonated with me, and I already had this work half-done in removing `#initialize` from the prepended module. 

Like I mentioned in my initial PR description, moving to a design that requires subclasses to implement `#handle` improves understandability of the library:

> When a subclass implements `#handle`, but the main interface remains `#call`, then application developers will implicitly understand that some extra code is being run in between the two.

This affordance is missing when the framework-provided behaviour is being prepended on the front of the application developer's behaviour. There's no hint that something special is going to happen (outside of reading docs).

Even though things were still _working_ with `#call` being in a prepended module, there were still some problems:

- Because the module prepending has to happen in an `.inherited` hook, it ends up being prepended multiple times
- This causes problems if you're trying to work with `Hanami::Action` subclasses in a deeper class hierarchy – it's not possible to call e.g. `super` inside the call method, because the prepended `#call` expects a single `env` argument, while the actual method in the concrete subclass takes `request, response` arguments.
- In this case, it adds a feeling of "magic" for no real benefit (unlike the work we still use to make it so the `configuration` option is applied "magically" during object instantiation, which does bring the benefit of making it so the concrete class' `#initialize` method doesn't need to deal with that (and can still call `super` however it wants).

I think the latter point is really the crux of it. If we can, we should aim as much as possible for objects in a Hanami app to work and feel like _regular ruby objects_. While `#call` still had prepended, this was not the case here.

Here's a couple of quick tests I wrote to demonstrate the problems. It may not be what you'd expect people to do in real apps, but IMO if we can design in such a way that people _can_ do things like this, all the better:

```ruby
module InheritanceTest
  class BaseAction < Hanami::Action
    def handle(*, res)
      res[:called_in_base] = true
    end
  end

  class LeafAction < BaseAction
    def handle(*, res)
      super
      res[:called_in_leaf] = true
    end
  end
end

RSpec.describe "Action inheritance" do
  subject(:action) { InheritanceTest::LeafAction.new(configuration: configuration) }

  let(:env) { Hash[] }

  it "adds framework-level behaviour only once" do
    expect(action.class.ancestors.select { |a| a == Hanami::Action::InstanceMethods }.length).to eq 1
  end

  it "supports typical class inheritance behavior across multiple inheriting action classes" do
    expect(action.call(env).exposures.keys).to include([:called_in_base, :called_in_leaf])
  end
end
```

Both of these tests pass with the latest code in this PR. Both will fail if we're still prepending behaviour.

The "unstable" branch of hanami-controller is already introducing major breaking changes, so I feel like now's the right time to introduce the requirement of this differently-named method.

Thoughts? Does this extra explanation help?

jodosha:
@timriley Thanks for the detailed explanation. The more "magic" we remove, the more we're happy. 

What I really want to keep is `#call` as public method to be overwritten in by concrete classes. We're moving in a direction of having _callable objects_, and `#handle` breaks this contract.

timriley:
@jodosha I'm not sure in this case that we can have our cake and eat it too.

If we keep `#call` as the method overwritten in the concrete classes, then we have to have "magic", i.e. the prepending of framework level behaviour, different params signatures for the different incarnations of `#call`, all of which will create confusion for application developers and also break the ability to leverage typical class inheritance behaviour. I think we'll also be stuck with the multiple prepending of the same module for deeper class hierarchies (which seems unavoidable unless we try for some great hacks).

If we use a method like `#handle` for the concrete classes, then these are _still_ "callable objects", but it is much clearer that the _library_ is providing the "callable" aspect, since the concrete class implements its own behaviour via a differently named method. No magic, less confusion. In fact, this follows the [template method pattern](https://en.wikipedia.org/wiki/Template_method_pattern), so there is prior art that people would be familiar with (just as solnic mentioned, this is exactly how rom-rb command class customisation happens, with command interface being `#call`, but the custom subclasses implementing `#execute)`. And along with all of this, application developers can leverage more elements of class inheritance behaviour like normal.

_Either way_, application developers are going to need to read the docs to see how to build a `Hanami::Action` subclass. I think it'll be far easier to explain the `#call`/`#handle` duality than it would be to explain the prepended behaviour.

AMHOL:
Also worth noting (although obvious that you meant something else when using the term contract) that the object is still callable, so although from the developers' implementation point of view, the contract is broken, the objects' interface still remains the same (`#call`).

One way to overcome this would be to have Hanami wrap actions when resolving them for the router, using a decorator to add the additional behaviour around the `#call` method, this still allows for the implementation to define the `#call(request, response)` interface, but leaves the objects half-implemented when instantiated directly, which could cause issues with existing tests that rely on the additional behaviour, feels like swings and roundabouts to me.

PR #277, opened 2018-11-16
Allow Hanami::Action to be inherited multiple times

jodosha:
Ensure actions to be able to:

  * Properly inherit callbacks chain from ancestors
  * Add callbacks without interfering with siblings and ancestors

This fix is partially based on @timriley work on `Action.new`.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -171,16 +171,25 @@ module Hanami
     # @since 0.1.0
     # @api private
     def self.inherited(base)
-      base.class_eval do
-        include Utils::ClassAttribute
-        class_attribute :before_callbacks
-        self.before_callbacks = Utils::Callbacks::Chain.new
+      if base.superclass == Hanami::Action
+        base.class_eval do
+          include Utils::ClassAttribute
 
-        class_attribute :after_callbacks
-        self.after_callbacks = Utils::Callbacks::Chain.new
+          class_attribute :before_callbacks
+          self.before_callbacks = Utils::Callbacks::Chain.new
 
-        prepend InstanceMethods
-        include Validatable if defined?(Validatable)
+          class_attribute :after_callbacks
+          self.after_callbacks = Utils::Callbacks::Chain.new
+
+          prepend InstanceMethods
+          include Validatable if defined?(Validatable)
+        end
+      else
+        base.class_eval do
+          self.before_callbacks = superclass.before_callbacks.dup
+          self.after_callbacks  = superclass.after_callbacks.dup
+          prepend InstanceMethods
+        end
       end
     end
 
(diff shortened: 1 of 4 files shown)


lib/hanami/action.rb
```diff
@@ -410,21 +419,25 @@ def self.handle_exception(exception)
...
+        obj.instance_variable_set(:@accepted_mime_types, Mime.restrict_mime_types(configuration, accepted_formats))
+        obj.instance_variable_set(
+          :@handled_exceptions,
+          Hash[
+            configuration
+            .handled_exceptions
+            .merge(handled_exceptions)
+            .sort{ |(ex1,_),(ex2,_)| ex1.ancestors.include?(ex2) ? -1 : 1 }
+          ]
+        )
```
timriley: I'd love to tidy this section up in a follow-up PR, if you're open to it, @jodosha :)

lib/hanami/action.rb
```diff
@@ -171,16 +171,25 @@ class Action
     # @since 0.1.0
     # @api private
     def self.inherited(base)
-      base.class_eval do
-        include Utils::ClassAttribute
-        class_attribute :before_callbacks
-        self.before_callbacks = Utils::Callbacks::Chain.new
+      if base.superclass == Hanami::Action
```
timriley: Could this be `if base.superclass = self` to eliminate the hard-coded name?

lib/hanami/action.rb
```diff
@@ -171,16 +171,25 @@ class Action
...
+          class_attribute :after_callbacks
+          self.after_callbacks = Utils::Callbacks::Chain.new
+
+          prepend InstanceMethods
+          include Validatable if defined?(Validatable)
+        end
+      else
+        base.class_eval do
+          self.before_callbacks = superclass.before_callbacks.dup
+          self.after_callbacks  = superclass.after_callbacks.dup
```
timriley: I wonder if we could make class attributes _always_ duplicate on inherited rather than having to manually arrange something like this in every class that uses them?

lib/hanami/action.rb
jodosha: I've got tricked by this too 😉 . In this context `self` is the superclass that you're inheriting from. For instance, if you're inheriting from `Web::Action`, then `self` is `Web::Action`.

In this case, I want only this behavior to be applied to **direct** descendants of `Hanami::Action`. By changing `Hanami::Action` with `self` causes spec failures.

lib/hanami/action.rb
jodosha: Yes, the entire gem needs a cleanup soon. But I'd punt on this for a little bit more.

From early benchmarks, this gem is slow 😢. We may want first to address these problems and then cleanup.

lib/hanami/action.rb
jodosha: Good point 👍I checked and this should happen by default: https://github.com/hanami/utils/blob/master/lib/hanami/utils/class_attribute.rb#L80

Let me investigate to see why it doesn't happen here..

timriley:
Ah yes! 👍🏼

lib/hanami/action.rb
jodosha: Actually, I pointed `master`, but should've been use this link https://github.com/hanami/utils/blob/unstable/lib/hanami/utils/class_attribute.rb#L94

jodosha: I turned out that these two lines were redundant. Thanks again @timriley! See https://github.com/hanami/controller/pull/277/commits/d5a829ee75345e9960522484a30977fe4342b5b3

jodosha:
@timriley Regarding `dry-auto_inject`, I'd not bring in another gem dependency for only this purpose. 😉

lib/hanami/action.rb
timriley: Great!

lib/hanami/action.rb
timriley: Ah yes, makes sense.

lib/hanami/action.rb
timriley: No probs, happy to leave this for later.

timriley:
> Regarding dry-auto_inject, I'd not bring in another gem dependency for only this purpose. 😉

Of course not. I just wanted to get it mentioned on the related PR, so it's on the record for those following from home :)

We'll maintain a spec in dry-auto_inject to ensure this new behaviour continues to work.

PR #279, opened 2018-12-16
remove flash and session from action

GustavoCaso:
Looks like from some merge conflicts or something similar we have some leftover code in `action.rb`.

Both methods are present in `response.rb`

@jodosha I wasn't sure to delete `MissingSessionError` class from the code, maybe we want to use it in the future to implement the missing functionality of raising it if the `Session` module is missing for both `response` and `request`

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -679,28 +679,6 @@ module Hanami
       end
     end
 
-    # Raise error when `Hanami::Action::Session` isn't included.
-    #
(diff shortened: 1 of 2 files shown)


tak1n:
I think we can close this PR, `Hanami::Action#flash` and `Hanami::Action#session` was removed in https://github.com/hanami/controller/pull/326.

timriley:
Ah yes, thanks for the triage, @tak1n!

PR #286, opened 2019-02-14
Introduce `Hanami::Action#name` to return a unique action name

jodosha:
## Enhancement

Return a serialized name for the action.
This is useful for `hanami` to know which action responded to an incoming request. The action name is set into the returned response.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -410,8 +410,10 @@ module Hanami
     end
 
     module Name
+      MODULE_SEPARATOR_TRANSFORMER = [:gsub, "::", "."].freeze
+
       def self.call(name)
-        Utils::String.transform(name, [:gsub, "::", "."], :underscore) unless name.nil?
+        Utils::String.transform(name, MODULE_SEPARATOR_TRANSFORMER, :underscore) unless name.nil?
       end
 
       class << self
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -11,6 +11,7 @@ require 'hanami/action/rack/file'
 
 require 'rack/utils'
 require 'hanami/utils'
+require 'hanami/utils/string'
 require 'hanami/utils/kernel'
 
 require 'hanami/utils/class_attribute'
@@ -408,12 +409,23 @@ module Hanami
       handled_exceptions.merge!(exception)
     end
 
+    module Name
+      def self.call(name)
+        Utils::String.transform(name, [:gsub, "::", "."], :underscore) unless name.nil?
+      end
+
+      class << self
+        alias [] call
+      end
+    end
+
     # Callbacks API instance methods
     #
     # @since 0.1.0
     # @api private
     def self.new(configuration:, **args)
       allocate.tap do |obj|
+        obj.instance_variable_set(:@name, Name[name])
         obj.instance_variable_set(:@configuration, configuration)
         obj.instance_variable_set(:@accepted_mime_types, Mime.restrict_mime_types(configuration, accepted_formats))
         obj.instance_variable_set(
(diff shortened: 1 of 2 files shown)


timriley:
Hi @jodosha, I can certainly see how this could make for a _tidier_-looking name than the `self.class.name` that was being assigned previously, but I'm not sure _why_ we actually need to be making this change? Would you mind explaining a little further what your reasoning was?

Also, I grepped through the codebase and can't actually see _where_ this name is being used (outside of it being assigned when a `Response` is created). Would you mind just pointing me in the right direction so I can understand the full extent of its usage a little better?

jodosha:
@timriley It's used on two places:

  1. At the boot time, to [register](https://github.com/hanami/hanami/pull/989/files#diff-1ed1deb125747568d4b8c969549060e5R96) the instance of an action in the container.
  1. At the runtime, to know which action was used to respond to the incoming HTTP request and to [pair the action with a view](https://github.com/hanami/hanami/pull/989/files#diff-62435de60db95dfeb2d3f4a112e5f80eR24).

timriley:
Thanks for that, @jodosha, that extra context helps a lot.

Now, if the _only_ use-case here was point 1, about needing to determine an identifier for registering actions within the container, then my suggestion would be that `Action.name` shouldn’t need to be changed (or perhaps could be removed altogether, maybe), because registration is concern that is _external_ to Action classes. And in this case, we could just as easily do something like this when registering them:

```ruby
Hanami::Utils::FileList[c.root, "apps", app.to_s, "actions", "**", "*.rb"].each do |path|
    action = loader.call(app, path, configuration)
    action_identifier = "..." # <- some path transformation logic to get action name, e.g. `books.index`

    register(:"apps.#{action_identifier}", action) unless action.nil?
  end

  # ...
end
```

(And maybe at some point we might even change the registration strategy to use dry-system auto-registration, or some variant of it, and having the determination of the container identifiers closer to the registration code could be handy at that point).

But of course there’s still the issue of the 2nd use case you pointed to, the one of determining a view to use when rendering an action’s response. It feels to me like this is an area that’s likely to change as we find a nice way to fit together stateless actions and views (particularly the kind of view objects that dry-view provides)... as well as considering ways to offer easy object composition via the container (e.g. via auto-injection or similar). If we think about it along these lines, perhaps a matching view object could be injected automatically at the time of the action object registration within the container (i.e. perhaps even in the general area of that code snippet above), which again would mean that the linking of an action to a matching view is handled by the container registration logic, and there needn’t be anything _intrinsic_ to the action that needs to know about a container identifier or anything like that.

Does that make sense?

So anyway, best path forward for this? Perhaps we merge this and accept it as a necessary interim step to keep everything working, but flag it as something we’ll definitely want to revisit when it comes time to refine our container setup and sort out the best approach for

jodosha:
@timriley Thanks for your comments. Let me reply one by one.

> Now, if the _only_ use-case here was point 1

I forgot to mention the standalone usage of `hanami-controller`. In a production API app based on it, I need to know which action responded to an incoming request. That is a useful information for me.

> And maybe at some point we might even change the registration strategy to use dry-system auto-registration

I'd happy to do so, but IMO auto-registration is a second step, after manual registration is in place. We're in an exploring phase, where manual is easier to change. In other words, with manual registration we can understand the requirements, and eventually adjust either `hanami` or `dry-container` code to create the perfect match.

> If we think about it along these lines, perhaps a matching view object could be injected automatically at the time of the action object registration within the container.

We can't assume that of one action there will be only a corresponding view. Since a long time, I was thinking if it's the case to implement `Hanami::Action#render`. Despite the fact it's a good idea or not, if we don't have a name structure for views, we would have hard times to match something like `render "books/new"`.

Another important aspect of action/view/repositories registration in a container is testing code. We can use `dry-system` _test mode_ to stub them, when an user is unit testing a component and wants to mock one of the core objects from the framework.

> So anyway, best path forward for this?

Yes, my suggestion is to merge it, complete the action/view registration first iteration, and eventually change/remove the code here.

PR #289, opened 2019-03-08
Do not error when If-Modified-Since is set but no validation exists for it

ianks:
We encountered a bug where an upstream proxy was setting the `Last-Modified` header automatically, which caused hanami controller to error with a 500.

It went something like this:

1. We want to cache on an etag, so we had the directive `fresh etag: 'some-value'`
2. Upstream proxy automatically adds the `Last-Modified` header
3. Browser sends `If-Modified-Since`
4. Since there is no validation value for `last_modified` validation was specified in the `fresh` call, `@value` is nil and causes hanami controller to error out

This PR fixes that issue by ensuring that `@value` is not blank.

### DIFF
--- a/lib/hanami/action/cache/conditional_get.rb
+++ b/lib/hanami/action/cache/conditional_get.rb
@@ -67,7 +67,10 @@ module Hanami
         # @since 0.3.0
         # @api private
         def fresh?
-          !Hanami::Utils::Blank.blank?(modified_since) && Time.httpdate(modified_since).to_i >= @value.to_time.to_i
+          return false if Hanami::Utils::Blank.blank?(modified_since)
+          return false if Hanami::Utils::Blank.blank?(@value)
+
+          Time.httpdate(modified_since).to_i >= @value.to_time.to_i
         end
 
         # @since 0.3.0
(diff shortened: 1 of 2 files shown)


lib/hanami/action/cache/conditional_get.rb
```diff
@@ -67,7 +67,10 @@ def initialize(env, value)
         # @since 0.3.0
         # @api private
         def fresh?
-          !Hanami::Utils::Blank.blank?(modified_since) && Time.httpdate(modified_since).to_i >= @value.to_time.to_i
+          return false if Hanami::Utils::Blank.blank?(modified_since)
+          return false if Hanami::Utils::Blank.blank?(@value)
```
davydovanton: Hey, not sure that I'm understand correctly your tests but could you add test case for this line too?

ianks: This is covered by my specs, if you remove the diff you'll notice it 500s without this line

davydovanton: Awesome! Thanks for explanation.

ianks:
Can this be merged?

ianks:
Just following back here. We would love to get this merged. In the meantime, we are doing this, which is unideal, but may help someone with the same issue.

```ruby
before do
  # remove once https://github.com/hanami/controller/pull/289 is merged
  request.env.delete Hanami::Action::Cache::IF_MODIFIED_SINCE

  # nginx will strip etags when gzip is requested unless they are weak
  fresh etag: "W/\"#{commit_sha}\""
end
```

Issue #285, opened 2019-02-09
Flash message appears twice (redux)


greggilbert:
Hey. Following #196, #260, etc., I'm having the double flash issue again in Hanami 1.3.1 (also 1.3.0).

```ruby
module Web
  class Failure
    include Web::Action

    def call(params)
      # shortened for brevity
      msg = "Something went wrong, try again!"
      flash[:message] = msg
      redirect_to routes.login_path
    end
  end
end
```

```ruby
module Web
  class Index
    include Web::Action
    include Hanami::Action::Session

    expose :message

    def call(params)

      message = flash[:message]
      if message
        @message = message
      end
    end
  end
end
```

As with some of the previous tickets, adding a `before { flash.clear }` fixes it. Am I not doing something right here?

jodosha:
@GustavoCaso Would you mind to check if this is a flash regression? Thanks.

jodosha:
@greggilbert What's the previous Hanami version you were on?

GustavoCaso:
@jodosha Of course. I will find some time this week to have a look at it.

@greggilbert would be able to provide a copy of your Gemfile?

PR #293, opened 2019-06-27
Don't persist flash after first redirect

jodosha:
Fixes https://github.com/hanami/controller/issues/285

### DIFF
--- a/spec/integration/hanami/controller/full_stack_spec.rb
+++ b/spec/integration/hanami/controller/full_stack_spec.rb
@@ -96,4 +96,15 @@ RSpec.describe "Full stack application" do
     expect(last_response.status).to be(302)
     expect(last_response.body).to   eq("Found") # This message is 302 status
   end
+
+  # See https://github.com/hanami/controller/issues/285
+  it "doesn't persist flash after first redirect" do
+    get "/failure"
+    follow_redirect!
+    follow_redirect!
+
+    expect(last_response.status).to be(200)
+    expect(last_response.body).to_not match("Ouch, access denied")
+    expect(last_response.body).to match("flash message not found")
+  end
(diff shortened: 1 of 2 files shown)


jodosha:
There is something that doesn't look good. The spec doesn't pass locally, but it does over CI. My intent was to reproduce the bug only, so the build supposed to not pass..

jodosha:
Nevermind. CI are reporting a false positive: https://travis-ci.org/hanami/controller/jobs/551367840 The build is failing but the `script/ci` exits with `0`, which confuses CI.

GustavoCaso:
I'm going to have a look @jodosha @AlfonsoUceda

GustavoCaso:
The logic seems fine, every time we trigger a redirect we store the values in `kept` if there is any `_data` and after the action is over we call `clear` on the `flash` updating the count for `kept` values and expire the ones that have been there for 2 or more requests. 

The test triggers 2 redirects and finally renders `Destination` but we access the `flash` before clearing the session, so the kept values haven't been deleted yet if the test were to trigger one redirect more the test will pass.

GustavoCaso:
@jodosha I took the liberty to add a commit https://github.com/hanami/controller/pull/293/commits/344e7371641b93875a7021f829556f5fddff3132 commenting your failing spec and adding one that proves my theory about adding an extra redirect.

Just wanted to start the conversation and what is the desired behaviour for the users. A possible solution is to expire kept data instead when reaching the count of `2` https://github.com/hanami/controller/blob/344e7371641b93875a7021f829556f5fddff3132/lib/hanami/action/flash.rb#L197 we could expire it to the count of `1`, but I'm sure this will break other tests.

Thoughts?

PR #298, opened 2019-10-17
Move storing of errors to utility for reusing

lfidnl:
@jodosha hello!
This is one more attempt  of https://github.com/hanami/hanami/pull/1014 :) Sorry, but I'm not sure to made it correct and place this utility in right place. If it is true, please, correct me and I will make appropriate changes.

### DIFF
--- a/lib/hanami/action/throwable.rb
+++ b/lib/hanami/action/throwable.rb
@@ -1,5 +1,6 @@
 require 'hanami/utils/class_attribute'
 require 'hanami/http/status'
+require 'hanami/action/rack/error_processing'
 
 module Hanami
   module Action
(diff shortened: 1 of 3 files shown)


lib/hanami/action/throwable.rb
```diff
@@ -144,31 +128,14 @@ def _rescue
           begin
             yield
           rescue => exception
-            _reference_in_rack_errors(exception)
```
jodosha: I would keep this private method:

```ruby
def _reference_in_rack_errors(exception)
  return unless configuration.handled_exception?(exception)

  Hanami::Action::Rack::ErrorProcessing.save_error_in_rack_env(@_env, exception)
end
```

lfidnl: No problem! I'll keep this method

lfidnl:
@jodosha I made changes according to your comment

PR #300, opened 2019-11-19
Fix multi-threading error on Rails app

alex-lairan:
Using Hanami Controller withing a Rails app cause a memoize of Action.

This memoized action can be used on multiples thread that break params
integrity.

### DIFF
--- a/lib/hanami/action/callable.rb
+++ b/lib/hanami/action/callable.rb
@@ -71,17 +71,17 @@ module Hanami
           @params ||= {}
           @params[Thread.current.object_id] = self.class.params_class.new(@_env)
           super params
-          @params.delete(Thread.current.object_id)
         end
 
         finish
       end
 
-      private
-
       def params
+        @params ||= {}
         @params[Thread.current.object_id]
       end
+
+      private
 
       # Prepare the Rack response before the control is returned to the
       # webserver.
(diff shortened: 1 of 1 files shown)


alex-lairan:
I have a Rails app with hanami-router and hanami-controller.

When a route match to a controller, the action is called.
The problem is that the action is memoized (I don't know why, I didn't found where [hanami or rails]),
so, when you have two concurent requests, the old params is overwrited.

Here an example of action :

```
class SomeAction
  include Hanami::Action

  # Don't use params here, use the method params instead
  def call(_)
    sleep 1
    puts params
    # here, params should be abc on the 1st call, and 123 on the 2nd
    # But because of threads, this will be abc 2 times
  end
end
```
```
curl -L http://localhost:3000/route_to_action?params_1=abc &; curl -L http://localhost:3000/route_to_action?params_1=123 &
```

jodosha:
@alex-lairan how do you define the route in the router?

alex-lairan:
We are defining our routes like this :

```ruby
Rails.application.routes.draw do
  router = Hanami::Router.new do
    get '/', to: SomeAction
  end

  mount router => '/api'
end
```

jodosha:
@alex-lairan Please have a look at this `README` section: https://github.com/hanami/controller#thread-safety

PR #311, opened 2020-04-29
View rendering via response object

timriley:


timriley:
This feature has been merged via #314.

PR #318, opened 2020-06-19
Use class-based configuration on Hanami::Action

timriley:
This PR updates `Hanami::Action` to use class-based configuration. For example:

```ruby
  # also aliased as config
  configuration.default_request_format = :html
end

action = MyAction.new
action.(env)
```

This will make it possible for us to automatically integrate actions into a full Hanami app using the same approach we do for views.

`Hanami::Action.configuration` is an instance of `Hanami::Action::Configuration` (renamed from `Hanami::Controller::Configuration`), which now uses dry-configurable to provide the standard setting readers/writers/processors, while also retaining the small handful of custom methods that were already in place for working with the settings.

This configuration object at the class level is mutable, so that class authors can adjust configuration settings at any time.

The class-level configuration object is also inheritable, which is what will allow a "base" application action to determine a baseline config for all its subclasses.

When actions are initialized, the configuration object from the class is injected, captured as an instance variable, and frozen at that point.

It's this injection of the class-level configuration that allowed me to make this change without major changes to the test suite, since it means it also supports manual injection of another configuration object, which is what most of the tests are currently doing. However, I think it should be the job of a later PR to update those tests to reflect recommended usage of the class-based configuration.


lib/hanami/action/configuration.rb
```diff
@@ -84,10 +35,25 @@ def initialize(&blk)
...
+      # @since 0.2.0
+      # @api private
+      DEFAULT_FORMATS = {
+        'application/octet-stream' => :all,
+        '*/*'                      => :all,
+        'text/html'                => :html
+      }.freeze
+
+      # TODO: docs
+      setting :formats, DEFAULT_FORMATS.dup
```
solnic, line 55: @timriley `dup` will make it mutable again, was that your intention?

timriley: @solnic This was actually copied over from the previous configuration class, and yes, I think it's actually the desired outcome, because e.g. the `format` method (just below this setting definition) will add new format entries to this hash.

timriley:
@jodosha This is ready for review and merging when ready.

With this in place, I'll then be able to move onto auto-configuring actions within Hanami applications.

PR #319, opened 2020-07-02
Include `Hanami::Action::Response#exposures` in view rendering context

jodosha:
## Syntax preview

```
  def handle(*, res)
    res[:users] = repo.all
    res[:foo] = "X"
    res.render(view, foo: "Y")
  end
end
```

The `:users` exposure is added to the view args for rendering.

In case of overlapping names between _inline options_ passed to `Hanami::Action::Response#render` and `Hanami::Action::Response#exposures`, _inline options_ will take the precedence.

In the example above the view will receive `"Y"` as value for `:foo`.

### DIFF
--- a/lib/hanami/action/response.rb
+++ b/lib/hanami/action/response.rb
@@ -76,7 +76,7 @@ module Hanami
       end
 
       def render(view, **options)
-        self.body = view.(**view_options.(request, self), **options.merge(exposures)).to_str
+        self.body = view.(**view_options.(request, self), **exposures.merge(options)).to_str
       end
 
       def format=(args)
--- a/lib/hanami/action/response.rb
+++ b/lib/hanami/action/response.rb
@@ -76,7 +76,7 @@ module Hanami
       end
 
       def render(view, **options)
-        self.body = view.(**view_options.(request, self), **options).to_str
+        self.body = view.(**view_options.(request, self), **options.merge(exposures)).to_str
       end
 
       def format=(args)
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -102,7 +102,7 @@ module Hanami
         end
 
         def finish(req, res, halted)
-          res.render(view, **req.params, **res.exposures) if render?(res)
+          res.render(view, **req.params) if render?(res)
           super
         end
 
--- a/spec/integration/hanami/controller/application_action/view_rendering_spec.rb
+++ b/spec/integration/hanami/controller/application_action/view_rendering_spec.rb
@@ -34,7 +34,8 @@ RSpec.describe "View rendering in application actions", :application_integration
 
               def handle(req, res)
                 res[:job] = "Singer"
-                res.render view, name: req.params[:name]
+                res[:age] = 0
+                res.render view, name: req.params[:name], age: 51
               end
             end
           end
@@ -70,7 +71,7 @@ RSpec.describe "View rendering in application actions", :application_integration
         module Main
           module Views
             class TestView < Main::View
-              expose :name, :job
+              expose :name, :job, :age
             end
           end
         end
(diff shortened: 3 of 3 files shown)


lib/hanami/action/response.rb
```diff
@@ -72,7 +72,7 @@ def body=(str)
       end
 
       def render(view, **options)
-        self.body = view.(**view_options.(request, self), **options).to_str
+        self.body = view.(**view_options.(request, self), **options.merge(exposures)).to_str
```
timriley, line 75: Regarding your notes in the main PR description:

> I'm not sure if options.merge(exposures) is the best way to do, because it will override explicit options in case of keys clash.
>
> ```ruby
> res[:foo] = "bar"
> res.render(view, foo: "baz")
> ```
> 
> With the current implementation :foo will return "bar" value. Do we care about this? How shall we solve it?

Could we change this to `**exposures.merge(options)`?

This way the exposures will be the implicit "base" for the options passed to the view render, and any _explicit_ options – the actual arguments passed to the view's `#call` – will take precedence, overriding the exposures in the case of matching keys, which feels like the right order to me.

What do you think @jodosha?

timriley:
@jodosha nice addition! I've left a comment with a suggestion that I think should improve the clarity of handling clashing exposures/view render options.

If you agree on that, it'd be nice to get this in.

Longer term, I think we want to have a good think/discussion about the approach we _recommend_ for application authors to follow re: passing objects to views and make sure that's well reflected in our future documentation. With both response exposures _and_ view exposures existing, I can see things getting pretty confusing.

lib/hanami/action/response.rb
jodosha: @timriley Thanks I had the same idea, happy that we're on the same page. 😃

jodosha: Fixed by 6c53dd97e4ebbf3fa0403e4a7050637876c40986

cllns:
After merging #345, I realized this PR provides the same functionality. It does it in a different place and higher up the call stack.

Which way is better @jodosha? Happy to use this one and revert #345. My understanding is that this PR would handle cases of people using `hanami-controller` with `hanami-action` with*out* using `ApplicationAction` (which is what's used for `hanami/hanami`.) and #345 would not handle that case.

jodosha:
@cllns @timriley Two things. ✌️ Please let me know if you agree.
@cllns In case of agreement, would you be able to expand this PR?

**One:** I would say it's better to push down to generic Action, even outside of `hanami`.

---

**Two:** We should consider the following usages:

**_Response exposures_**

We set an exposure in the response, which should be available to the corresponding view/template. This is what #345 is taking care of.

```ruby
  def handle(*, res)
    res[:foo] = "bar"
  end
end
```

**_Rendering exposures_**

We explicitly set an exposure while invoking `res.render`. `:foo` must be available in the view/template. This is what this PR is taking care of. Dunno if this PR also covers #345 as well.

```ruby
  def handle(*, res)
    res.render(view, foo: "bar")
  end
end
```

**_A mix of the previous use cases_**

Both `:color` and `:foo` must be available in the view/template. In case of conflict, the `Hanami::Reponse#render` wins.

```ruby
  def handle(*, res)
    res[:color] = "red"
    res.render(view, foo: "bar")
  end
end
```

timriley:
@jodosha I agree with both your suggestions, and would love to see this PR expanded to cover them! @cllns, let me know what you think and if you'd be happy to work on this 😄

cllns:
@timriley @jodosha Yup, I can see this PR through

cllns:
Rebased this branch, removed the change from #345 and all the tests still pass 🙌 . So this original change to `Hanami::Action::Response#render` does handle the actual sending the exposures to the view and we don't need to do it twice. Now it's at the appropriate spot, since it'll work for both `StandaloneAction`s and `ApplicationAction`s.

@jodosha @timriley please review

jodosha:
@cllns LGTM 👍 I'm the original author of the PR, so I can't approve it 😉

timriley (approved):
This is great 👍🏼

PR #322, opened 2020-07-22
Auto-inject inferred paired view into application actions

timriley:
Update the `ApplicationAction` behavior to automatically auto-inject a "paired" view into any action, based on the action name. This behavior is provided by a configurable `ViewNameInferrer`.

For example, for an action named `Main::Actions::Articles::Index`, inject the view found at the `"views.articles.index"` registration.

It will not attempt to inject the view if no such registration exists.

This auto-injected view is also manually overridable for testing purposes, etc., i.e. `Actions::Articles::Index.new(view: my_custom_view)`.

Special behavior is also provided for RESTful actions named `Create` or `Update`. In these cases, the view name inferrer will return two view identifiers: `["views.articles.create", "views.articles.new"]`, where the second identifier is an "alternative" view, matching the expected view that would have led to the given RESTful request ("create"->"new" and "update"->"edit"). This is useful because it will allow the application author to easily re-render the matching view with error messages in the case of an invalid record.

In this case, where two potential view names are provided by the inferrer, the action will attempt to look them up in order, so in the example above, if a view is _actually_ registered at `"views.articles.create"`, then that will be provided instead of `"views.articles.new"`.

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -3,34 +3,23 @@
 module Hanami
   class Action
     class ApplicationAction < Module
-      DynamicInstanceMethods = Class.new(Module)
-
       attr_reader :provider
       attr_reader :application
-      attr_reader :instance_mod
 
       def initialize(provider)
         @provider = provider
         @application = provider.respond_to?(:application) ? provider.application : Hanami.application
-        @instance_mod = DynamicInstanceMethods.new
       end
 
       def included(action_class)
-        # TODO: I think we should probably do:
-        #
-        #   instance_mod.include StaticInstanceMethods
-        #
-        # And just include the single instance_mod, rather than these two
-        # modules, for a simpler ancestors chain
-        action_class.include StaticInstanceMethods
-        action_class.include instance_mod
+        action_class.include InstanceMethods
 
         define_initialize action_class
         configure_action action_class
       end
 
       def inspect
-        "#<#{self.class.name}[#{provider}]>"
+        "#<#{self.class.name}[#{provider.name}]>"
       end
 
       private
@@ -74,7 +63,7 @@ module Hanami
         end
       end
 
-      module StaticInstanceMethods
+      module InstanceMethods
         # FIXME: Can I turn these into attr_readers?
         def view
           @view
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -30,8 +30,11 @@ module Hanami
 
         define_method :initialize do |**deps|
           super(**deps)
-          @view = deps[:view] || resolve_view.(action_class)
-          @view_context = deps[:view_context] || resolve_context.()
+
+          # Conditionally assign these to repsect any explictly auto-injected
+          # dependencies provided by the class
+          @view ||= deps[:view] || resolve_view.(action_class)
+          @view_context ||= deps[:view_context] || resolve_context.()
         end
       end
 
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -54,7 +54,7 @@ module Hanami
           provider: provider
         )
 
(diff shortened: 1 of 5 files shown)


lib/hanami/action/application_action.rb
```diff
@@ -51,6 +67,11 @@ def configure_action(action_class)
       end
 
       module InstanceMethods
+        # FIXME: Can I turn these into attr_readers?
```
solnic, line 70: @timriley yes, what was your concern?

lib/hanami/action/view_name_inferrer.rb
```diff
@@ -0,0 +1,46 @@
...
+  class Action
+    class ViewNameInferrer
+      ALTERNATIVE_NAMES = {
+        "create" => "new",
+        "update" => "edit"
+      }.freeze
+
+      class << self
+        def call(action_name:, provider:)
+          application = provider.respond_to?(:application) ? provider.application : Hanami.application
```
solnic, line 13: @timriley why is this needed?

lib/hanami/action/application_action.rb
timriley: Mostly a "note to self" that I never got around to adjusting. I'll do this now :)

lib/hanami/action/view_name_inferrer.rb
timriley: @solnic Because it's possible for the `provider` to be the `Hanami.application` itself, not just a slice. And at the moment `Hanami::Application` does not offer an `.application` method. So we need this to support actions that are in both the applications's root `/lib/` as well as within slices.

I have a card to create an `Hanami::Application.application` method (simply returning `self`) so we no longer need to do these `provider.respond_to?` checks: https://trello.com/c/Y7JWzo7y

lib/hanami/action/application_action.rb
timriley: Fixed :)

lib/hanami/action/application_action.rb
```diff
@@ -43,6 +48,17 @@ def resolve_view_context
         end
       end
 
+      def resolve_paired_view(action_class)
+        view_identifiers = application.config.actions.view_name_inferrer.(
+          action_name: action_class.name,
+          provider: provider
+        )
+
+        view_identifiers.each_with_object(nil) { |identifier|
```
jodosha, line 57: @timriley very very minor thing: why this and not `view_identifiers.find`? What am I missing?

spec/integration/hanami/controller/application_action/view_rendering/paired_view_inference_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        expect(action.view).to be view
+      end
+
+      context "Another view explicitly auto-injected" do
+        let(:action_class) {
+          module Main
+            module Actions
+              module Articles
+                class Index < Hanami::Action
+                  include Deps[view: "views.articles.custom"]
```
jodosha, line 48: @timriley 👏

spec/integration/hanami/controller/application_action/view_rendering/paired_view_inference_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+      Main::Actions::Articles::Create
+    }
+
+    let(:direct_paired_view) { double(:direct_paired_view) }
+    let(:alternative_paired_view) { double(:alternative_paired_view) }
+
+    context "Direct paired view exists" do
+      before do
+        Main::Slice.register "views.articles.create", direct_paired_view
```
jodosha, line 93: @timriley 💯

lib/hanami/action/application_action.rb
timriley: @jodosha You're right! This could with with `#find`/`#detect` too. I'm going to switch it over.

PR #324, opened 2020-07-30
Render view automatically within application actions

timriley:
This removes the last piece of boilerplate for basic application actions: if a view is available (either through paired view inference or explicit auto-injection), it will be called, with all the request parameters passed to it.

Along with the paired view inference, this means we can now create an action like this:

```ruby
module Main
  module Actions
    module Articles
      class Index < Hanami::Action
      end
    end
  end
end
```

And it will automatically render the `Main::Views::Articles::Index`!

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -70,6 +70,14 @@ module Hanami
         attr_reader :view
         attr_reader :view_context
 
+        protected
+
+        def handle(request, response)
+          if view
+            response.render view, **request.params
+          end
+        end
+
         private
 
         def build_response(**options)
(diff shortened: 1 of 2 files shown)


lib/hanami/action/application_action.rb
```diff
@@ -70,6 +70,14 @@ module InstanceMethods
         attr_reader :view
         attr_reader :view_context
 
+        protected
+
+        def handle(request, response)
+          if respond_to?(:view) && view
```
solnic, line 76: @timriley why does it check `respond_to?(:view)`? I see `attr_reader :view` on line 71 so...every view responds to this method, no?

timriley: @solnic you're right, this wasn't needed, thanks for checking! Removing it now.

PR #327, opened 2020-09-21
Reintroduce CSRFProtection module

timriley:
This module disappeared from the hanami repo during the beginnings of the 2.0 overhaul. Adding it back to hanami-controller makes sense, since this is where the CSRF protection will be taking place.

This introduces the module (and some corresponding shiny new unit tests), but does not yet automatically enable it if sessions are enabled. I'll do the automatic enablement in a follow-up PR.


timriley:
[Restore CSRF protection behavior](https://trello.com/c/nPpmQPRd/56-restore-csrf-protection-behavior)

jodosha (approved):

PR #329, opened 2020-09-29
Enable CSRF protection automatically

timriley:
Enabled CSRF protection automatically on application actions if the application config has `sessions.enabled?`.

Provide a `config.actions.csrf_protection` setting to allow the application author to explicitly opt out of this if they want (by setting it to false).

This behaviour relies on the change in https://github.com/hanami/hanami/pull/1074, which passes the application config object directly to the nested actions config, so it can inspect it and find out whether it has `sessions.enabled?`. Doing it this way felt like a nicer approach than e.g. passing the `sessions` config directly, or making another `sessions_enabled` setting on the actions config and setting that directly from the outside, because both of those feel too specific to this one feature, rather than something that can be generalisable to any 3rd-party provided config objects.

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -16,6 +16,7 @@ module Hanami
 
         define_initialize action_class
         configure_action action_class
+        extend_behavior action_class
       end
 
       def inspect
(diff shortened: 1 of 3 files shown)


timriley:
Closing in favour of #332

PR #331, opened 2020-09-30
Forward all args from `.new` to `#initialize`

### LINKED ISSUE #330: Standalone actions prevent using non-keyword parameters in `#initialize` (v2)
I just got tripped up by this:

```ruby
  def initialize(arg)
    @arg = arg
  end

  def handle(*)
  end
end

MyAction.new(5)
# !!! ArgumentError: wrong number of arguments (given 1, expected 0)
#     from [...]/lib/hanami/action/standalone_action.rb:262:in `new'
```

The exception message is confusing because it sounds like `.new` expected zero arguments.

This is caused by `StandaloneAction.new` only forwarding keyword arguments to `#initialize`. Could it forward positional and block args too? I tested it quickly and it seemed ok.

Let me know if you want a PR.

tomdalling:
Previously only forwarded keyword arguments. Now also forwards
positional and block arguments.

Closes #330

### DIFF
--- a/lib/hanami/action/standalone_action.rb
+++ b/lib/hanami/action/standalone_action.rb
@@ -249,22 +249,22 @@ module Hanami
 
         # Returns a new action
         #
-        # @overload new(**args)
-        #   @param args [Hash] action dependencies
+        # @overload new(**deps, ...)
+        #   @param deps [Hash] action dependencies
         #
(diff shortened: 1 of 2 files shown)


timriley:
Thanks @tomdalling!

PR #332, opened 2020-09-30
Enable CSRF protection automatically

timriley:
Enable CSRF protection automatically on application actions if the application config has `sessions.enabled?`.

What it does is:

- Move the `sessions` config into hanami-controller, so it is accessible via `config.actions.sessions` (as part of `ApplicationConfiguration`)
- Adds a `csrf_protection` setting to `ApplicationConfiguration`, defaulting to `nil`
- Uses `ApplicationConfiguration#finalize!` to set `csrf_protection` to true or false based on whether sessions are enabled (but _only_ if it was previously `nil`, which allows the application author to explicitly disable it if they desire)

This relies upon https://github.com/hanami/hanami/pull/1078, which calls `finalize!` on the `config.actions` and `config.views` objects.

In addition to this, it also makes it possible to use _all_ the public methods on methods on `Hanami::Action::Configuration` when accessing it via an `ApplicationConfiguration`, i.e. via `Hanami.application.config.actions`. To support this, we now delegate to `Hanami::Action::Configuration` as a "base configuration" from within `Hanami::Action::ApplicationConfiguration`, which means all methods work now, not just the defined settings.

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -16,6 +16,7 @@ module Hanami
 
         define_initialize action_class
         configure_action action_class
+        extend_behavior action_class
       end
 
       def inspect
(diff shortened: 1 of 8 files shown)


timriley:
[Restore CSRF protection behavior](https://trello.com/c/nPpmQPRd/56-restore-csrf-protection-behavior)

timriley:
[Custom methods on `Hanami::Action::Configuration` are not available when accessed as `Hanami.application.config.actions`](https://trello.com/c/CbkZD8cd/114-custom-methods-on-hanamiactionconfiguration-are-not-available-when-accessed-as-hanamiapplicationconfigactions)

jodosha (approved):

PR #333, opened 2020-10-07
Remove Action#with

timriley:
This was introduced to make actions compatible with the previous incarnation of Hanami::Router. With the latest changes to the router, it is no longer required.


timriley:
[Remove Action#with](https://trello.com/c/iH8ATE5r/115-remove-actionwith)

jodosha (approved):
@timriley Thanks for taking care of this.

PR #334, opened 2020-10-16
Fix unstable CI for Ruby 2.6

jodosha:


timriley:
@jodosha Thank you! Since this is such a simple fix, I'll merge it right away :)

PR #336, opened 2020-10-31
Set default security-related headers for application actions

timriley:
When Actions are used within the context of a full Hanami application, by default they will now serve a range of security-related headers by default (`X-Frame-Options`, `X-Content-Type-Options`, `X-XSS-Protection`, `Content-Security-Policy`). This matches the behaviour established in Hanami 1.x.

Previously, these headers were independently configurable via the application's `config.security` namespace, e.g. `config.security.x_frame_options = "SAMEORIGIN"`. Instead of following this approach, we've instead just configured the existing `default_headers` hash with these header values automatically. This is a simpler, more maintainable approach, and just as easy to understand to users, who would already be familiar with the header values they're looking to configure.


timriley:
[Work out where/how we should be using the `cookies` and `security` application settings](https://trello.com/c/LcC64ra5/89-work-out-where-how-we-should-be-using-the-cookies-and-security-application-settings)

jodosha (approved):
@timriley Thanks. I manually tested it to verify that these default headers aren't sent when an action is used in standalone mode. 👍

PR #337, opened 2020-10-31
Enable cookie support in application actions, based on config

timriley:
For application actions, include the `Action::Cookies` module automatically if cookies are enabled on the application-level actions config (`config.actions.cookies.enabled?`). Along with this, ensure the application-level configured cookie options still properly pass through to the action.

Allow automatic cookie support to be disabled by setting `config.actions.cookies = nil`.

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -72,6 +72,11 @@ module Hanami
           require "hanami/action/csrf_protection"
           action_class.include Hanami::Action::CSRFProtection
         end
+
+        if application.config.actions.cookies.enabled?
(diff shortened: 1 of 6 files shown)


timriley:
[Work out where/how we should be using the `cookies` and `security` application settings](https://trello.com/c/LcC64ra5/89-work-out-where-how-we-should-be-using-the-cookies-and-security-application-settings)

jodosha (approved):
@timriley Thanks for this new piece of the puzzle. 👍

jodosha:
@timriley I merged `unstable` here because this should fix CI for MRI 2.6.

PR #338, opened 2021-01-13
What should go here to configure Hanami router?

radar:
If we're using `hanami-router` 2.0.0.alpha3 + `hanami-controller` 2.0.0.alpha1 and attempt to use the syntax as suggested in hanami-controller's README, we see this error:

```
ArgumentError:
  unknown keyword: :configuration
```

We can define routes without this keyword:

```ruby
module Twist
  module Web
    Router = Hanami::Router.new do
      if ENV['APP_ENV'] == "development"
        require 'sidekiq/web'
        mount Sidekiq::Web, at: '/sidekiq'
      end

      post '/graphql', to: Controllers::Graphql::Run.new
      options '/graphql', to: Controllers::Graphql::Run.new

      post '/books/:permalink/receive', to: Controllers::Books::Receive.new

      get '/oauth/authorize', to: Controllers::Oauth::Authorize.new
      get '/oauth/callback', to: Controllers::Oauth::Callback.new
    end
  end
end
```

However, each `new` call here apparently needs a `configuration` keyword:

```
ArgumentError:
  missing keyword: :configuration
# ./lib/twist_web/router.rb:9:in `block in <module:Web>'
```

---

Looks like if I use the `unstable` branches of both `hanami-controller` + `hanami-router`, I can get away without providing a `configuration` key (using the above router configuration). 

So perhaps this PR should be to remove the `configuration` and possibly also the `namespace` options from the `Hanami::Router` call?

### DIFF
--- a/README.md
+++ b/README.md
@@ -895,7 +895,7 @@ module Web
 end
 
 configuration = Hanami::Controller::Configuration.new
-router = Hanami::Router.new(configuration: configuration, namespace: Web::Controllers) do
+router = Hanami::Router.new(???, namespace: Web::Controllers) do
   get "/books/:id", "books#show"
 end
 ```


jodosha (changes requested):
@radar The documentation is outdated.

If you're using `hanami-controller` + `hanami-router` only, you can't reference actions by a string key anymore (e.g. `books#show`).

To achieve that feature, `hanami-router` used to act as an implicit container which was responsible to translate keys (`books#show`) into the appropriate class name (via inflection), require the file (knowing framework conventions), reference the class (via reflection), and instantiate it using the right action configuration. **That was too much for a router.**

**You can still use this convenient syntax with `hanami`**, but for _standalone usage_, like in your case, the arrangement is more DIY.

The correct syntax is:

```ruby
configuration = Hanami::Controller::Configuration.new

Hanami::Router.new do
  get "/books/:id", Books::Show.new(configuration: configuration)
end
```

PR #339, opened 2021-02-03
Support ruby 3.0

hieuk09:
In Ruby 3.0, positional arguments and keyword arguments will be separated.

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -210,10 +210,18 @@ module Hanami
       module InstanceMethods
         # @since 0.7.0
         # @api private
-        def initialize(*)
-          super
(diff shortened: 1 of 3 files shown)


adam12:
Thank you for the patch.

Unfortunately, supporting Ruby 3 on the 1.x lineage of `hanami-controller` will likely be incredibly difficult due to some of it's dependencies and _their_ dependencies.

The `unstable` branch is tracking the 2.x lineage, which we hope to have out soon and will support Ruby 3.

I'm going to close this, as while the changes look benign enough, we will not be able to test them in CI without considerable more work. We may choose to re-evaluate and/or re-open this PR but at this point, our efforts are going towards Hanami 2.

PR #345, opened 2021-05-14
Pass exposures to view (in addition to params)

cllns:
We should automatically send all exposures (which are definfed with `response[:foo] = :bar`) to the view, when one exists. 

If there are most tests to add for this, let me know and I'd be happy to add them :)


timriley (approved):
Thanks @cllns! I'd still like revisit how exposures from actions work (I haven't spent much time thinking about it yet), especially given we now have a same-named feature of the view layer, but in the meantime, I think this is a really rational change, and we shouldn't intentionally hobble the action exposures while they still exist in their current form :)

jodosha (approved):
@cllns Thanks. It makes totally sense (I thought it was already there 😉 ).

PR #346, opened 2021-05-20
Move automatic view rendering to `finish` (from `handle`)

cllns:
The way `Hanami::Action::ApplicationAction` works right now (which is automatically included in `Hanami::Action`s when in a Hanami application), users will have to remember to include `super` everytime they override `handle` in order to get the automatic view rendering. Implementing that method is the main way people will add functionality to their actions, so forgetting to include a `super` seems like it'll happen a lot and be frustrating for users.

```ruby
module Main
  module Actions
    module Home
      class Index < Main::Action
        def handle(req, res)
          res[:foo] = :bar
          super # This sets res.body with the associated view rendering. If this is removed, nothing is rendered.
        end
      end
    end
  end
end
```

After this PR, that rendering for the view automatically happens (as long as they don't set a `response.body` manually, or `halt` the flow).

```ruby
module Main
  module Actions
    module Home
      class Index < Main::Action
        def handle(req, res)
          res[:foo] = :bar
          # The associated view is rendered and stored in res.body, as long as it's not manually set here.
        end
      end
    end
  end
end
```
This lets users avoid having to write `super` in every implementation of `handle` for their actions.

**NOTE**: This PR is based off the branch for #345. This base branch will need to change to `unstable` if/when that's merged (which should happen first). Wanted to avoid merge conflicts. Happy to rebase this against `unstable` if we want this and not that PR though.

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -87,7 +87,7 @@ module Hanami
 
         def handle(request, response)
           if view
-            response.render view, **request.params
+            response.render(view, **request.params, **response.exposures)
           end
         end
 
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -83,14 +83,6 @@ module Hanami
         attr_reader :view
         attr_reader :view_context
 
-        protected
-
-        def handle(request, response)
-          if view
-            response.render(view, **request.params, **response.exposures)
-          end
-        end
-
         private
 
         def build_response(**options)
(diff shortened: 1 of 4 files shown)


timriley (approved):
This is a fantastic improvement, thank you @cllns!

lib/hanami/action/application_action.rb
```diff
@@ -105,6 +97,12 @@ def view_options(req, res)
         def view_context_options(req, res)
           {request: req, response: res}
         end
+
+        # Automatically render the view, if the body hasn't been populated yet
+        def finish(req, res, halted)
+          res.render(view, **req.params, **res.exposures) if view && res.body.empty?
```
jodosha, line 103: @cllns As next step, I would extract a _public API_ method:

```ruby
def render?(res)
  view && res.body.empty?
end
```

Pros:

  * Clarity of code
  * Let final users to override this method

jodosha (approved):
@cllns LGTM. 👍  It makes sense. Left a comment about a refactoring opportunity.

PR #349, opened 2021-08-09
When configuration public directory isn't set, fallback to default value

jodosha:
https://github.com/hanami/hanami-2-application-template/commit/9a23a1b331dc0ee7629099d0d6320a3f341e4885 [crashes](https://gist.github.com/jodosha/99254d58c7e7432830184dc5d4bc5085) when I try to run the specs:

---

The [current implementation](https://github.com/hanami/controller/blob/d59d5b6eda2fa5229f4291bb503c031abd31899e/lib/hanami/action/configuration.rb#L408-L411) of `Hanami::Action::Configuration#public_directory` is:

```ruby
def public_directory
  root_directory.join(super).to_s
end
```

For the App Template `super` returns `nil`, which makes the `Pathname#join` (`root_directory` is an instance of `Pathname`) to crash.

---

This fix aims to fallback to the default value if `super` returns `nil`.

---

**EDIT: ⚠️ I'm not sure this is the right approach. Given the default value is assigned via `dry-configurable`, I would have expected `super` to return the default value, instead of `nil`.**

### DIFF
--- a/lib/hanami/action/configuration.rb
+++ b/lib/hanami/action/configuration.rb
@@ -407,7 +407,7 @@ module Hanami
       # @see root_directory=
       def public_directory
         # This must be a string, for Rack compatibility
-        root_directory.join(super).to_s
+        root_directory.join(super || DEFAULT_PUBLIC_DIRECTORY).to_s
       end
 
       private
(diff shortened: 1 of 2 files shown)


mereghost:
If you set if to `nil` `dry-configurable` will return `nil` even if there's a default as it assumes that was intentional (IIRC - has been a while since I delved into ts code). Can you share a sample where this became an issue?

I've removed the assignment and everything works as expected.

waiting-for-dev:
> For the App Template super returns nil.

Where is it set on the App Template? I looked into both the App Template and Hanami itself, and I couldn't find it.

PR #353, opened 2021-11-10
Introduce `Hanami::Action::ApplicationConfiguration#content_security_policy`

jodosha:
## Feature

Provide a human friendly API to interact with Content Security Policy settings.

## Reasoning

Content Security Policy (CSP) is a default HTTP response header.
We have a feature that handles default headers, but giving that CSP is a large blob of text, if a dev needed to alter just a bit, they need to rewrite the entire CSP header. That wasn't human friendly.

This PR introduces an easier API to deal with CSP, by splitting the several settings into a Hash manipulation experience.

### Read a setting

```ruby
module MyApp
  class Application < Hanami::Application
    config.actions.content_security_policy[:base_uri] # => "'self'"
  end
end
```

### Override a setting

```ruby
module MyApp
  class Application < Hanami::Application
    # This line will generate the following CSP fragment
    # plugin-types ;
    config.actions.content_security_policy[:plugin_types] = nil
  end
end
```

### Append to a default value

```ruby
module MyApp
  class Application < Hanami::Application
    # This line will generate the following CSP fragment
    # script-src 'self' https://my.cdn.test;
    config.actions.content_security_policy[:script_src] += " https://my.cdn.test"
  end
end
```

### Add a custom setting

```ruby
module MyApp
  class Application < Hanami::Application
    # This line will generate the following CSP fragment
    # my-custom-setting 'self';
    config.actions.content_security_policy[:my-custom-setting] = "'self'"
  end
end
```

### Delete a setting

```ruby
module MyApp
  class Application < Hanami::Application
    config.actions.content_security_policy.delete(:object_src)
  end
end
```

### Disable CSP

```ruby
module MyApp
  class Application < Hanami::Application
    config.actions.content_security_policy = false
  end
end
```

### DIFF
--- a/lib/hanami/action/application_configuration/content_security_policy.rb
+++ b/lib/hanami/action/application_configuration/content_security_policy.rb
@@ -9,7 +9,7 @@ module Hanami
       class ContentSecurityPolicy
         # @since 2.0.0
         # @api private
-        def initialize
+        def initialize(&blk)
           @policy = {
             base_uri: "'self'",
             child_src: "'self'",
@@ -26,6 +26,15 @@ module Hanami
             script_src: "'self'",
             style_src: "'self' 'unsafe-inline' https:"
           }
+
+          blk&.(self)
+        end
+
+        # @since 2.0.0
+        # @api private
+        def initialize_copy(original_object)
+          @policy = original_object.instance_variable_get(:@policy).dup
+          super
         end
 
         # Get a CSP setting
--- a/lib/hanami/action/application_configuration/content_security_policy.rb
+++ b/lib/hanami/action/application_configuration/content_security_policy.rb
@@ -52,7 +52,7 @@ module Hanami
         #   end
         # end
         def [](key)
-          @policy.fetch(key, nil)
+          @policy[key]
         end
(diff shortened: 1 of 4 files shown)


lib/hanami/action/application_configuration/content_security_policy.rb
```diff
@@ -0,0 +1,101 @@
...
+        # @api public
+        #
+        # @example
+        # module MyApp
+        #   class Application < Hanami::Application
+        #     config.actions.content_security_policy[:base_uri] # => "'self'"
+        #   end
+        # end
+        def [](key)
+          @policy.fetch(key, nil)
```
mereghost, line 55: Wouldn't this be the same than `@policy[key]` since we are defaulting to nil?

jodosha: @mereghost You're right. This is a leftover.

jodosha:
```suggestion
          @policy[key]
```

mereghost (approved):

PR #354, opened 2021-11-12
Refactor/standardize spec file paths

cllns:
With https://github.com/hanami/hanami/pull/1123 we use compact source directories. 

The specs work either way, but we should be consistent


jodosha (approved):

timriley:
Thanks @cllns!

PR #355, opened 2021-11-12
Extract Hanami::Action::ApplicationAction#render?

cllns:
See https://github.com/hanami/controller/pull/346/files\#r744831907Implementing the work suggested in this comment: https://github.com/hanami/controller/pull/346/files#r744831907

We don't have a way to document that it's a part of the public api (e.g. `# @api public`) yet. I expect we'll go through during a later pass and do that then. It would be strange to document just this one method with YARD

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -101,10 +101,14 @@ module Hanami
           {request: req, response: res}
         end
 
-        # Automatically render the view, if the body hasn't been populated yet
         def finish(req, res, halted)
-          res.render(view, **req.params, **res.exposures) if view && res.body.empty?
+          res.render(view, **req.params, **res.exposures) if render?(res)
           super
+        end
+
+        # This can be overridden to enable/disable automatic rendering
+        def render?(res)
+          view && res.body.empty?
         end
       end
     end
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -106,7 +106,15 @@ module Hanami
           super
         end
 
-        # This can be overridden to enable/disable automatic rendering
+        # Decide whether to render the current response with the associated view.
+        # This can be overridden to enable/disable automatic rendering.
+        #
+        # @param response [Hanami::Action::Response]
+        #
+        # @return [TrueClass,FalseClass]
+        #
+        # @since 2.0.0
+        # @api public
         def render?(res)
           view && res.body.empty?
         end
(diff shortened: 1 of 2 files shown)


lib/hanami/action/application_action.rb
```diff
@@ -101,11 +101,15 @@ def view_context_options(req, res)
...
         end
 
-        # Automatically render the view, if the body hasn't been populated yet
         def finish(req, res, halted)
-          res.render(view, **req.params, **res.exposures) if view && res.body.empty?
+          res.render(view, **req.params, **res.exposures) if render?(res)
           super
         end
+
+        # This can be overridden to enable/disable automatic rendering
```
jodosha, line 109:
```suggestion
        # Decide if render the current response with the associated view.
        # This can be overridden to enable/disable automatic rendering.
        #
        # @param [Hanami::Action::Response]
        # @return [TrueClass,FalseClass]
        #
        # @since 2.0.0
        # @api public
```

jodosha (approved):
@cllns LGTM 👍

jodosha:
@cllns Now I noticed this sentence:

> We don't have a way to document that it's a part of the public api (e.g. # @api public) yet.

Why?

cllns:
@jodosha Ah sorry, I just meant there weren't other YARD comments in the file yet.

I added a spec for this method

spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb
```diff
@@ -42,6 +42,50 @@ class Test < Main::View
...
+
+              def render?(_res)
+                false
+              end
+            end
+          end
+        end
+      RUBY
+
+      write "slices/main/lib/views/test.rb", <<~RUBY
```
jodosha, line 65: @cllns This has `lib` in the path.

Should we try to verify that the action works event without a view and template?

cllns: I think we already test that later in the file, without `render?` coming into play at all. I can add an example with `render?` being `false` if you want, but it seems redundant to me

https://github.com/hanami/controller/blob/4dfd319e38a9c6c93b8469ed1708e13d1f2b4335/spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb#L171-L191

cllns: Good call about lib :+1: I just merged #354, rebased, and removed the `lib` from the path in these specs

jodosha: @cllns OK, if that is already tested, we're good.

PR #356, opened 2021-11-12
Raise error for missing view

cllns:
Before:
```ruby
view = nil
response.render(view)
# #<NoMethodError: undefined method `call' for nil:NilClass>
```

After
```ruby
view = nil
response.render(view)
#<Hanami::Controller::MissingViewError: Cannot render a view that is missing>
```

We should provide helpful errors to users when possible.

(This came about from implementing the spec for #355)

Happy to change the wording

### DIFF
--- a/lib/hanami/controller/error.rb
+++ b/lib/hanami/controller/error.rb
@@ -11,8 +11,8 @@ module Hanami
     #
     # @since 2.0.0
     class MissingViewError < Error
-      def initialize
-        super("Cannot render a view that is missing")
+      def initialize(klass)
+        super("missing view for #{klass}")
       end
     end
   end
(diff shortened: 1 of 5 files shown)


lib/hanami/controller/error.rb
```diff
@@ -3,5 +3,17 @@ module Controller
...
+
+    # Missing view error.
+    #
+    # It's raised when an ApplicationAction should automatically render a view,
+    # but there's no view associated with the instance
+    #
+    # @since 2.0.0
+    class MissingViewError < Error
+      def initialize
+        super("Cannot render a view that is missing")
```
jodosha, line 15:
```suggestion
        super("missing view for #{klass}")
```

jodosha (changes requested):

cllns:
Thanks @jodosha. I just implemented the line you asked for, though I'm not sure it's what you were looking for. I feel like it might make sense to surface the _where_ the Action looked for views, rather than just the name of the class (which will show up in the stacktrace anyway).

Also happy to re-implement the spec in a different way, stubbing out `class` as a String isn't terrific.

PR #357, opened 2021-11-23
Add ApplicationAction#routes to access Hanami app routes

waiting-for-dev:
References hanami/hanami#1119

### DIFF
--- a/spec/integration/hanami/controller/application_action/routes_spec.rb
+++ b/spec/integration/hanami/controller/application_action/routes_spec.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+require "hanami"
+require "hanami/action"
+
+RSpec.describe "Routes", :application_integration do
+  specify "Access application routes from an action" do
+    with_tmp_directory(Dir.mktmpdir) do
+      write "config/application.rb", <<~RUBY
+        require "hanami"
+
+        module TestApp
+          class Application < Hanami::Application
+          end
+        end
+      RUBY
+
+      write "config/routes.rb", <<~RUBY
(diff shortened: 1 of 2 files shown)


spec/integration/hanami/controller/application_action/routes_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+          end
+        end
+      RUBY
+
+      write "slices/main/actions/test_action.rb", <<~RUBY
+        module Main
+          module Actions
+            class TestAction < Hanami::Action
+              def handle(req, res)
+                res.body = "Hello, world!"
```
jodosha, line 35: @waiting-for-dev I would take the chance of testing the fact that those helpers are accessible here.

```suggestion
                res.body = routes.root_url
```

spec/integration/hanami/controller/application_action/routes_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+            end
+          end
+        end
+      RUBY
+
+      require "hanami/init"
+
+      action = Main::Slice["actions.test_action"]
+
+      expect(action.routes.path(:root)).to eq("/")
```
jodosha, line 46: @waiting-for-dev I would perform a request and assert that the body of the response is equal to the root URL.

jodosha (changes requested):
@waiting-for-dev Thanks for this PR.

I believe we should use Dependency Injection here.
Preferred way is to import via `dry-system` (`include Deps["routes"]`).

waiting-for-dev:
Closing in favor of #358

PR #358, opened 2021-11-24
Provide access to routes inside all application actions

timriley:
This builds upon @waiting-for-dev's work in #357 and makes the routes an injectable dependency of application actions.

This needed a complementary change in Hanami (https://github.com/hanami/hanami/pull/1132) to make it so the Hanami application's endpoint resolver doesn't eagerly initialize an action instance for each of the endpoints. If it continued to do that, with this change here in this PR, we run into a deadlock error around accessing/loading the router inside `Hanami::Application.router`.

This reason for this can hopefully be made clear through stepping through the sequence of fetching the first action from the container (like we do in the test added in #357):

1. An  action is resolved from a (non-booted) Hanami application's slice
2. The action's `#initialize` will resolve the `router_helper` from the application container
3. The router_helper boot file will get the `Hanami.application.router`
4. This begins to load the router, which, using the application's endpoint resolver, will eagerly create an action instance for each of the declared routes
5. The first action instance has an `#initialize` that will resolve the `router_helper` from the application's container
6. And we're back at step 2 again, so we're now in an endless loop

As for the changes inside this PR in particular, I think it's the right way to do it. The routes helper _should_ be an injectable dependency, defaulting to the framework-provided helper if its available.

As for the changes to the Hanami router initialization and behaviour, there's likely a whole variety of different approaches we can take to make it complement the change here, and be more flexible in general. I think there's a lot to be said for _not_ eagerly initializing every action whenever you load the router, because it makes Hanami more suited to different deployment modes (e.g. deployed to a Lambda for a subset of routes only; in this case it is a waste of resources and cold boot time to initialize every action when you're not necessarily going to use them all).

To put it another way, one of the key features we offer in Hanami 2 is giving the user maximal control over their application boot, and I think we should make sure our router doesn't compromise this.

I think the change I've done in https://github.com/hanami/hanami/pull/1132 is a reasonable starting point, however, and we can consider more comprehensive changes at a later moment.

What do you think, @waiting-for-dev, @jodosha?

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -14,7 +14,7 @@ module Hanami
       def included(action_class)
         action_class.include InstanceMethods
 
-        define_initialize action_class
+        define_initialize
         configure_action action_class
         extend_behavior action_class
       end
@@ -25,7 +25,7 @@ module Hanami
 
       private
 
-      def define_initialize(action_class)
+      def define_initialize
         resolve_view = method(:resolve_paired_view)
         resolve_context = method(:resolve_view_context)
         resolve_routes = method(:resolve_routes)
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -106,6 +106,16 @@ module Hanami
           super
         end
 
+        # Access application routes
+        #
+        # @return [Hanami::Application::RoutesHelper]
+        #
+        # @since 2.0.0
+        # @see Hanami::Application::RoutesHelper
+        def routes
+          Hanami.application[:routes_helper]
+        end
+
         # Decide whether to render the current response with the associated view.
         # This can be overridden to enable/disable automatic rendering.
         #
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -28,24 +28,16 @@ module Hanami
       def define_initialize(action_class)
         resolve_view = method(:resolve_paired_view)
         resolve_context = method(:resolve_view_context)
+        resolve_routes = method(:resolve_routes)
 
         define_method :initialize do |**deps|
           # Conditionally assign these to repsect any explictly auto-injected
           # dependencies provided by the class
           @view ||= deps[:view] || resolve_view.(self.class)
           @view_context ||= deps[:view_context] || resolve_context.()
+          @routes ||= deps[:routes] || resolve_routes.()
 
           super(**deps)
-        end
-      end
-
-      def resolve_view_context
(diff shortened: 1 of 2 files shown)


jodosha:
@timriley Yes totally agreed and thanks for the changes.

---

👇 Long explanation ahead. 👇 
The changes that I did in `hanami-router` from 1 to 2 removed the responsibility from the router to load the endpoints.

The reasons were:

1. The coupling with the rest of Hanami conventions
2. The undermined reusability of the router

This shift from 1 to 2 made `hanami-router` more a general purpose library.
On the other hand that responsibility of loading the endpoints now must be handled by `hanami`.
Which I believe it's coherent with the framework philosophy.

Before this PR (and the associated one), booting an application would mean to boot the router and by consequence all the referenced actions. This is a waste of resources (and time) in development mode.

Making the endpoint resolver lazy (as you did in https://github.com/hanami/hanami/pull/1132) doesn't just solve the problem of the infinite loop, but also makes the application faster to restart in development mode.

We have a great win-win situation here! 💯

lib/hanami/action/application_action.rb
```diff
@@ -49,15 +62,8 @@ def resolve_view_context
...
-        view_identifiers = application.config.actions.view_name_inferrer.(
-          action_name: action_class.name,
-          provider: provider
-        )
-
-        view_identifiers.detect { |identifier|
-          break provider[identifier] if provider.key?(identifier)
-        }
+      def resolve_routes
+        application[:routes_helper] if application.key?(:routes_helper)
```
waiting-for-dev, line 66: Isn't it defensive checking that `:router_helper` exists? Shouldn't it always exist as we're registering it from `hanami/hanami`?

waiting-for-dev:
Thanks, @timriley! I first tried with the same approach, but I found the deadlock error. With your change on `hanami/hanami` the world makes sense again :slightly_smiling_face:

lib/hanami/action/application_action.rb
timriley: Good pick up :) At the moment, it's over-defensive, I agree! However, I would like at some point to move all the router integration opt-in/out like the rest of the system, which means we could very well have applications without a routes_helper. I'd much rather have this do the right thing from the beginning rather than have to be changed later. Does that make sense?

timriley:
Glad you're happy with the approach here, @jodosha and @waiting-for-dev!

I've just made these PRs "ready for review" — let me know if you're good for me to merge them both as-is.

jodosha (approved):

timriley:
Resolves #357.

PR #359, opened 2021-12-02
Do not render view when application action halts with a body

ianks:
Previously, halting inside of an application would not actually prevent a view from rendering. This patch fixes that by making sure to set the response status and body if a request was halted. 

cc @timriley

### DIFF
--- a/spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb
+++ b/spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb
@@ -216,7 +216,7 @@ RSpec.describe "Application actions / View rendering / Automatic rendering", :ap
       RUBY
 
       write "slices/main/web/templates/test.html.slim", <<~'SLIM'
-        #{raise("Oh no"}
+        == raise("This should not be rendered")
       SLIM
 
       require "hanami/init"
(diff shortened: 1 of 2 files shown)


cllns:
Why doesn't `super` handle this? Is it because `ApplicationAction` and `StandaloneAction` are mixins, and they're out of order? It'd be great if we could find some way to leverage that, rather than duplicating this line https://github.com/hanami/controller/blob/edaf5ebfb5218f3ca3bdb3982657a3f189b7b350/lib/hanami/action/standalone_action.rb#L566

spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb
```diff
@@ -190,6 +190,45 @@ class Test < Hanami::Action
...
+        module Main
+          module Views
+            class Test < Main::View
+            end
+          end
+        end
+      RUBY
+
+      write "slices/main/web/templates/test.html.slim", <<~'SLIM'
+        #{raise("Oh no"}
```
timriley, line 219: Is this _intended_ to be invalid syntax?

timriley:
Agree with @cllns that it would be good if we could leverage the ancestors and the existing behaviour in `StandaloneAction` to handle this. @ianks, would you mind having a go at that?

ianks:
@timriley @cllns Made that change 😄

timriley (approved):
Looks good, thanks!

@jodosha Worth getting this in before the release this week too?

timriley:
n.b. We've decided not to rush this into this month's release (happening in the next day or so). We'll defer until next month's release so we have some more time to consider the solution.

tak1n:
Thank you @ianks for your contribution. This was somewhat forgotten and has been fixed by another [PR](https://github.com/hanami/controller/pull/372) (Sorry for that)

PR #360, opened 2021-12-13
Add RSS MIME type to TYPES

parndt:
When trying to use RSS I noticed that the content type wasn't listed as
valid in the `TYPES` hash.


timriley (approved):
Thanks Phil!

jodosha (approved):

PR #364, opened 2022-02-02
Update HTTP Status codes

cllns:
I figure we can start fresh for Hanami 2, following Rack's list exactly at this moment in time.


jodosha (approved):
@cllns Good one 👍

PR #365, opened 2022-02-03
[Proposal] Change ApplicationAction to a class

cllns:
**TIP:** When viewing the diff in "Files changes", select "Hide whitespace" from the gear icon, for a much cleaner diff.

Right now, without this PR, inheriting from `Hanami::Action` automagically integrates with views/templates, view context and routes if it was done inside a `Hanami::Application`. This is done by including an `ApplicationAction` module. For more information on this, see [Tim's blog post](https://timriley.info/writing/2020/06/28/open-source-status-update-june-2020/) about implementing it.

With this proposed change, users needing Application integration in their Actions would instead inherit from `Hanami::ApplicationAction`, rather than `Hanami::Action`.

While this is _slightly_ less convenient, this is largely mitigated by Hanami 2 leveraging inheritance for project and slice-specific superclasses, so for almost all users, they'll only have `Hanami::ApplicationAction` written **once** in their codebase. 

### Why switch?
**A**. It's more maintainable. Less meta-programming (though there is still some in the configuration). It's easier to understand. The existing code is hard to reason about and hard to change. Actions are, in a sense, the core of a web framework, so this is especially valuable. The code is run for every request in a Hanami::Application. If/when users run into bugs, they can look inside the framework code and follow it pretty easily. This also encourages people to contribute :)

**B**. We're also able to leverage good object-oriented design, class inheritance and dependency injection with keyword args. Though libraries are different from apps, being able to model what we think is "good code" is a benefit.

**C**. We can rely on more unit tests instead of integration tests. 🤩 I was getting order-dependent integration spec failures during developing this work. They seemed to have gone away, but more unit tests instead of integration tests will make maintenance easier. (It seems like when _one_ spec fails, others non-deterministically fail. Maybe due to skipping some teardown?)

**D**. It's explicit. It also lets users opt-out of `Hanami::Application` integration for some actions/slices. Without this change, (I think?) the view/routes integration automatically happens and there's no obvious way to disable it, though we could obviously fix that (with more code).

### Further work
If we want to do this, a couple questions:

1. Should we move this file to `hanami/hanami` with the name `Hanami::Application::Action`? This follows with a convention of using inheritance to _specialize_ behavior in a certain use-case. This makes `hanami/hanami` just another implementer of the `Hanami::Action` API. Other people could use it to write their own custom framework actions. 
    
    I'm in favor of this. It respects our namespaces better, for one. 

    Note that this does add some complexity to development compared to now, if there are synced changes in `hanami/controller` and `hanami/hanami`.

    Another option could be keeping it here, but renaming it to `Hanami::Action::Application`. I think this is worse, but it is an option.

2. Re: the `StandaloneAction` module, which is where the actual behavior for `Hanami::Action` is implemented. I'd like to see either: 

    a. We remove it entirely and put all its code into `Hanami::Action` class. It kind of feels like we're 'hiding' code artificially, especially after this PR. We may be able to extract another class or two from it, to make the file smaller.

    b. We rename it. It was named `Callable` in Hanami 1, perhaps we should name it `Handleable`, since overriding the `handle` method is the main API for the objects. It's bit awkward for an adjective, but it's descriptive.

    Should we do either one?

3. Should we do the same for `hanami/view`? It has the same pattern with `StandaloneView` and `ApplicationView`. Should we do this there too? I say yes. They are different applications but being consistent is valuable, and I expect we'd get the same cost/benefit breakdown.

If we want to go down this path, here's some more work I'd like to do before merging:

- [ ] Convert many of the integration specs to unit specs
- [ ] Possible: move to `hanami/hanami`, as `Hanami::Application::Action`
- [ ] Possible: Rename `StandaloneAction`.
- [ ] Possible: Duplicate this work for `hanami/view` which uses the same pattern


timriley:
@cllns FYI, I'm currently using your work here as the beginnings of some experimentation about shifting this code into Hanami. Thanks for contributing it, I'll let you know how it goes :)

tak1n:
@cllns @timriley this can be closed right? From what I can tell this has landed as https://github.com/hanami/hanami/pull/1156 in `hanami/hanami`.

tak1n:
@cllns @timriley this can be closed right? From what I can tell this has landed as https://github.com/hanami/hanami/pull/1156 in `hanami/hanami`.

jodosha:
@cllns Thanks, but I'm closing this as stale, because we already implemented the new inheritance chain for H2.

PR #372, opened 2022-04-11
Do not automatically render halted requests.

andrewcroome:
When upgrading a hanami 2 application from the unstable branch to hanami 2.0.0.alpha6, I discovered that redirected requests are unexpectedly rendering their views (resulting in errors where variables in slim templates are not present and ultimately in failed redirects).

i.e. calling the following action with no name param results in the slim being rendered (and a `NoMethodError: undefined method 'upcase' for nil:NilClass`), despite the redirect_to being invoked:

```ruby
# slices/main/actions/profile/show.rb
module Main
  module Actions
    module Profile
      class Show < Action::Base
        before :redirect_unless_name

        def handle(req, res)
          res[:name] = req.params[:name]
        end

        private

        def redirect_unless_name(req, res)
          res.redirect_to "/" unless req.params[:name]
        end
      end
    end
  end
end

# slices/main/templates/profile/show.html.slim
h1 Hello #{name.upcase}
```

This PR adds a check to confirm that a request is not halted before automatically rendering its view. Without this check, the spec added here fails with `NoMethodError: undefined method 'upcase' for nil:NilClass`.

### DIFF
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -109,7 +109,7 @@ module Hanami
         end
 
         def finish(req, res, halted)
-          res.render(view, **req.params) if render?(res, halted)
+          res.render(view, **req.params) if !halted && render?(res)
           super
         end
 
@@ -122,8 +122,8 @@ module Hanami
         #
         # @since 2.0.0
         # @api public
-        def render?(res, halted)
-          view && res.body.empty? && halted.nil?
+        def render?(res)
+          view && res.body.empty?
         end
       end
     end
--- a/lib/hanami/action/application_action.rb
+++ b/lib/hanami/action/application_action.rb
@@ -109,7 +109,7 @@ module Hanami
         end
 
         def finish(req, res, halted)
-          res.render(view, **req.params) if render?(res)
+          res.render(view, **req.params) if render?(res, halted)
           super
         end
 
@@ -122,8 +122,8 @@ module Hanami
         #
         # @since 2.0.0
         # @api public
-        def render?(res)
-          view && res.body.empty?
+        def render?(res, halted)
+          view && res.body.empty? && halted.nil?
         end
       end
     end
--- a/spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb
+++ b/spec/integration/hanami/controller/application_action/view_rendering/automatic_rendering_spec.rb
@@ -55,7 +55,7 @@ RSpec.describe "Application actions / View rendering / Automatic rendering", :ap
                   res[:favorite_number] = 123
                 end
 
-                def render?(_res)
+                def render?(_res, _halted)
                   false
                 end
               end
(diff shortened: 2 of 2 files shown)


lib/hanami/action/application_action.rb
```diff
@@ -122,8 +122,8 @@ def finish(req, res, halted)
         #
         # @since 2.0.0
         # @api public
-        def render?(res)
-          view && res.body.empty?
+        def render?(res, halted)
+          view && res.body.empty? && halted.nil?
```
timriley, line 126: @andrewcroome What do you think about shifting the order of the conditionals around?

```suggestion
          !halted && view && res.body.empty?
```

If we know the action is already halted, feels like we should short circuit from that point, and not even bother with the rest of the checks.

timriley: Actually wait. There's another alternative, what do you think about this as a reader?

```ruby
        def finish(req, res, halted)
          res.render(view, **req.params) if !halted && render?(res)
          super
        end

        def render?(res)
          view && res.body.empty?
        end
```

This way we keep the `halted` logic contained to `#finish`, and keep `#render?` focus on the view-rendering conditions only.

andrewcroome: @timriley I think that second option feels good, yes. I've added a commit which adopts it.

timriley: For the record, I check in with @jodosha and he's happy for us to merge this code as is, which I'll do shortly 😄

timriley (approved):

timriley:
Thank you @andrewcroome!

timriley:
@andrewcroome Just FYI, this code will be fairly short-lived in the main branch in its current state, due to the work happening in https://github.com/hanami/hanami/pull/1156 to shift the action/view integration code into the Hanami gem itself.

I shifted your change into that branch in https://github.com/hanami/hanami/pull/1156/commits/fab689450a171b80e4d8b7cdfee00f98122f7377, and made a change to the tests at the same time...

What was bothering me was that we already had a similar test to yours, the `""Doesn't render view automatically when halt is called"` example. In poking around, I realised the difference:

- In the existing test, the view still rendered, but because we supplied all the necessary params for it to render successfully, that render wasn't apparent, because it was response's body was then overwritten by the standard `halted` handling in base action class’ `#finish`
- In your new test in this PR, you specifically did _not_ provide the necessary params for the view to render successfully, so that render would raise an exception, which is what you saw originally and what prompted you to make this change

What is common to both of those tests, however, is that they both halt.

So in my commit over in that other PR, what I did was update the existing test to incorporate some elements from yours, and added some comments explaining why those changes were in place (i.e. any attempt to render the view will crash, so the fact the test passes in the first place is evidence the view never rendered). This way we have just one test covering the meaningful behaviour here, rather than two with just incidental differences.

andrewcroome:
Great thanks @timriley and @jodosha. Yes, that existing spec was passing because the view being rendered happened to render happily, which hid the fact that the render was occurring. Agree that just having the one spec is better, and `to_str` is neater than my `upcase` + `redirect_to` example. Thank you for giving this your attentions!

PR #373, opened 2022-04-26
Fix gem name in README

okuramasafumi:


jodosha (approved):
@okuramasafumi Thank you!

PR #374, opened 2022-05-11
HEAD request fixes

narinda:
This PR addresses two issues with HEAD requests

### HEAD requests should return the same headers as GET requests without the message body
Currently HEAD requests are returning a message body which causes a `Rack::Lint` error in development. According to the RFC https://datatracker.ietf.org/doc/html/rfc2616#page-54, HEAD requests should return the same headers as the equivalent GET request. Currently headers are being filtered to exclude all but a subset of Entity headers, Content-type and Content-length are also being filtered. This has implications also for HEAD requests that return a redirect response. 

The method `_requires_empty_headers` included a check for HEAD requests. I've removed this and I've added a separate method `_empty_body` that also sets the message body to empty if the request is a HEAD request. 

### HEAD requests for redirects are not returning redirect headers 
The redirect_to method on the response cancels the redirect if there is already a body present or file being sent. This check is made using `#renderable?` which also includes a check for HEAD requests. This is causing requests that would return redirect headers on GET to instead try to render the request on a HEAD request. This has been causing errors in our app as those pages are often not renderable. I've created a new method called `allow_redirect?` that is used instead and retains the check for an existing message body or file but allows redirect headers to be sent with a HEAD request.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -515,9 +515,7 @@ module Hanami
 
     # @since 2.0.0
     # @api private
-    def _requires_empty_headers?(res)
-      _requires_no_body?(res) || res.head?
-    end
+    alias_method :_requires_empty_headers?, :_requires_no_body?
 
     private
 
@@ -669,6 +667,11 @@ module Hanami
       res.headers.select! { |header, _| keep_response_header?(header) }
     end
 
+    # @api private
+    def _empty_body(res)
+      res.body = Response::EMPTY_BODY
+    end
+
     def format(value)
       case value
       when Symbol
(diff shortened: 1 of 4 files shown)


timriley:
Thank you @narinda! I'll check this out tomorrow morning :)

timriley:
Welp, I just woke from a gooooood month-long sleep, and I've checked this first thing in the morning as promised 😉 

(Sorry for the delay, @narinda!)

All looks right to me. I'm going to merge this since I know you're waiting for it. It should make it into the next alpha release, which is currently slated for ~next week.

@jodosha, if you have any other thoughts, would you mind weighing in post-merge? Seems like our HEAD-handling was not fully compliant before... perhaps there might have been a reason for it?

PR #376, opened 2022-05-16
Developer friendly error for params class method

tak1n:
This will most likely not be a problem in a full-blown hanami application, as I could imagine the overall `hanami` gem would also install `hanami/validations`. But if using `hanami/controller` standalone this will raise a more developer-friendly error to instruct to install `hanami/validations`.

It is already somewhat documented in the README (explicit require for `hanami/validations`), but a more helpful runtime error could not harm.

Feel free to point out some way how to test this or whether the test can be skipped as it is such a minor thing.

EDIT: Found the isolation specs :)


lib/hanami/action.rb
```diff
@@ -214,6 +214,19 @@ def self.accepted_formats
...
+    #
+    # Raises a developer friendly error to include `hanami/validations`.
+    #
+    # @raise [NoMethodError]
+    #
+    # @api private
+    # @since 2.0.0
+    def self.params(klass = nil, &blk)
+      raise NoMethodError,
+            "To use Hanami::Action.params include hanami/validations gem in your Gemfile"
```
cllns, line 227:
```suggestion
            "To use `params`, please add 'hanami/validations' gem to your Gemfile"
```

cllns (commented):
Great idea! I made a minor suggested change to the language, what do you think of it?

tak1n:
> Great idea! I made a minor suggested change to the language, what do you think of it?

Sounds way better, thanks. ✌🏽

timriley:
Nice improvement, thank you!

PR #377, opened 2022-05-16
Prefer NoMethodError over NotImplementedError

tak1n:
See the discussion at https://github.com/rmosolgo/graphql-ruby/issues/2067

We could as well implement a custom error, like is done in graphql-ruby with `RequiredImplementationMissingError`. I used `NoMethodError` for now as this was already thrown for the missing params method in https://github.com/hanami/controller/blob/main/spec/isolation/without_hanami_validations_spec.rb#L23.

### DIFF
--- a/lib/hanami/action/request.rb
+++ b/lib/hanami/action/request.rb
@@ -43,52 +43,52 @@ module Hanami
         @accept ||= @env[HTTP_ACCEPT] || DEFAULT_ACCEPT
       end
 
-      # @raise [NotImplementedError]
+      # @raise [NoMethodError]
       #
       # @since 0.3.1
       # @api private
       def content_type
-        raise NotImplementedError, 'Please use Action#content_type'
+        raise NoMethodError, 'Please use Action#content_type'
       end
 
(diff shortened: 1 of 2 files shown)


timriley:
I'm in favour of this change

In the past I would also use `NotImplementedError` myself, event knowing it was intended for something slightly different, but I didn't realise that it didn't inherit from `StandardError`, so I agree it'd be a good change to make.

If we make this change here, however, I'd want us to make sure we do a sweep of the other Hanami projects to make sure we make similar changes for any other instances of `NotImplementedError`, too.

@tak1n — if you haven't done it already, would you mind checking the rest of our projects?

Before I merge this in, @jodosha — are you OK with this adjustment? And from your perspective, is this actually worth mentioning in the changelog as a "breaking change?" (My hunch is that it's unlikely users are actually intentionally catching these errors)

tak1n:
> @tak1n — if you haven't done it already, would you mind checking the rest of our projects?

Will take care of it.

jodosha (approved):
@tak1n Nice one, thanks! 👍

PR #378, opened 2022-05-17
Readd helpful errors when sessions are not enabled

tak1n:
Tracked by https://trello.com/c/MJtJHbcC/104-re-add-helpful-errors-when-session-is-accessed-off-response-but-sessions-are-not-enabled.

Basically [copy-pasted](https://github.com/hanami/controller/pull/326/files) the old error class, `#flash` and `#session` under the `Hanami::Action` namespace/class.


tak1n:
This is not ready, as `flash` and `session` is no longer accessed on the action itself, but rather on the request/response object.

tak1n:
Will open a new PR once ready.

Issue #256, opened 2018-04-16
Make `Hanami::Action::{Request,Response}` `#session` and `#flash` to raise `Hanami::Controller::MissingSessionError` when session is not configured


jodosha:
With https://github.com/hanami/controller/pull/250 (v1.2.0) we introduced a new exception (`Hanami::Controller::MissingSessionError`) which is raised when a dev tries to access `Action#session` and `Action#flash` when the HTTP session isn't configured.

In 2.0 (`unstable` branch), the access to `session` and `flash` will be moved from `Hanami::Action` to `Hanami::Action::Request` and `Response`. These two objects need to support the old behavior of raising a `Hanami::Controller::MissingSessionError`.

---

This change **IS** for `unstable` branch.

---

Ref https://github.com/hanami/controller/pull/250

GustavoCaso:
Sorry, that was a wrong click 😄

timriley:
Fixed in #379.

PR #379, opened 2022-05-18
Readd session errors

tak1n:
Somewhat unhappy with the implementation, but this introduces some tests as well on which we can rely on to refactor this. Feel free to suggest improvements, maybe we could leverage `Hanami::Action::Configuration` here. 

The configuration is already injected into the response object, we could take care to inject it into the request object as well.

Closes https://github.com/hanami/controller/issues/256

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -22,6 +22,7 @@ require_relative "action/mime"
 require_relative "action/rack/file"
 require_relative "action/request"
 require_relative "action/response"
+require_relative 'action/error'
 
 module Hanami
   # An HTTP endpoint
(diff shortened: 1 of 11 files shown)


tak1n:
[Re-add helpful errors when session is accessed off response, but sessions are not enabled](https://trello.com/c/MJtJHbcC/104-re-add-helpful-errors-when-session-is-accessed-off-response-but-sessions-are-not-enabled)

jodosha:
@tak1n Could you please rebase with `main`? 🙏

tak1n:
@jodosha done, also fixed remaining rubocop offenses.

timriley:
Fixed in #395.

PR #381, opened 2022-05-23
Return HTTP status 415 for requests with wrong Content-Type

tak1n:


tak1n:
[Action enforce_accepted_mime_types callback should return 415 status code upon failure, not 406](https://trello.com/c/oPeUNfC5/112-action-enforceacceptedmimetypes-callback-should-return-415-status-code-upon-failure-not-406)

jodosha (approved):
@tak1n LGTM, thanks! 👍

PR #384, opened 2022-07-10
Establish `Hanami::Action#call` as Public API

jodosha:
# Enhancement

Re-establish `Hanami::Action#call` as Public API for Hanami 2.0 actions.

During the Hanami 2.0 rewriting, we had to gave up to a _Hanami signature_ method `#call` for actions in favor of `#handle`. This was to reduce the internal complexity (meta-programming) of `Hanami::Action`.

# Implementation details

Introduce `Hanami::Action::Proxy`, a wrapper around `Hanami::Action` that acts as an adapter with Rack.
It's responsible to prepare the execution (instantiate request, response, run callbacks) and finally invokes `Hanami::Action#call`.

This wrapper eliminates the need of meta-programming to keep `Hanami::Action#call` as Public API.

For each incoming Rack request

```
[rack server] --> [router] --> *[action proxy] --> [action]

* new addition of this PR.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ They are the endpoints that respond to incoming HTTP requests.
 
 ```ruby
 class Show < Hanami::Action
-  def handle(req, res)
+  def call(req, res)
     res[:article] = ArticleRepository.new.find(req.params[:id])
   end
 end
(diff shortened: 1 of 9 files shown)


timriley:
Closing this one for now, since we decided we'll continue with `#handle(request, response)` as the key method to be implemented by users.

PR #392, opened 2022-09-26
Reduce memory usage in Hanami::Action via dry-configurable and custom config_class

timriley:
This PR takes advantage of the recent memory usage improvements in dry-configurable (see https://github.com/dry-rb/dry-configurable/pull/138) by converting the previously custom `Configuration` class into settings defined on the `Hanam::Action` class.

To preserve some of the existing custom configuration methods, we provide a custom config class to `Dry::Configurable` and keep those methods there.

This custom config class also comes in handy as the place to keep the documentation for our settings, which I've converted into yard `@!attribute` macros for concision.

We also move two existing bits of state into new dry-configurable settings:

- Convert `before_callbacks` and `after_callbacks` from class attributes, since Hanami-utils' class attributes implementation is very inefficient with memory.
- Move `accepted_formats` from being an class-level ivar, since this was the only remaining class-level state and it made sense to keep all this together via the class-level config object.

As a consequence of the last change, we're also able to change `Hanami::Action` so it has a conventional `#initialize` method again, rather than having a custom `.new` method that has to `allocate` the object and assign some ivars directly, which is a big win for ease of understanding the code :)

Lastly, while we're here, we get rid of the @name ivar for actions, since it's no longer used anywhere.

Using the newly added `benchmarks/memory_profile_action.rb` (which creates 100 `Action` subclasses by default) we can see this branch allocates 0.338 MB of memory, versus 1.859 MB for the main branch.

Even putting aside the memory improvements, I'd paint this as a useful refactor, because it reduces code, better leverages a library we're already using (dry-configurable), as well as centralises all the action class-level state, making it easier for us to manage.

🙋🏼 **I've tidied up the commit history here so you should be able to review things commit by commit.**


timriley:
FYI, it seems we had a test failure in a previous run of this (https://github.com/hanami/controller/actions/runs/3166800527/jobs/5156785405) — I'll look into this and address it before merging.

spec/integration/hanami/controller/sessions_spec.rb
```diff
@@ -58,7 +58,7 @@ def response
   include Rack::Test::Methods
 
   let(:app) do
-    Hanami::Action::Configuration.new
+    # Hanami::Action::Configuration.new
```
solnic, line 61: Is this needed here?

solnic (approved):
This is wonderful 👏🏻 I'm happy to see that the new config_class feature turned out to be useful here already 🙂

spec/integration/hanami/controller/sessions_spec.rb
timriley: Oh yeah, I can clean this up! Thanks for the note :)

(The fact that we no longer need to do this in 2.0 is excellent)

timriley:
@solnic It's so good! Couldn't have done it without you :)

timriley:
> FYI, it seems we had a test failure in a previous run of this (https://github.com/hanami/controller/actions/runs/3166800527/jobs/5156785405) — I'll look into this and address it before merging.

FYI, I looked at this test and couldn't replicate locally (same Ruby version, RSpec command and seed), plus it doesn't seem related to my changes here, so I'm not following it up any further at this point.

solnic:
> @solnic It's so good! Couldn't have done it without you 🙂

🥹

timriley:
[Update Hanami::Action to use class-level config and latest dry-configurable (reduce memory usage)](https://trello.com/c/ExtpYdCm/208-update-hanamiaction-to-use-class-level-config-and-latest-dry-configurable-reduce-memory-usage)

PR #394, opened 2022-10-14
Provide consistent class-level interface for config-like methods

### LINKED ISSUE #389: Handling `config` and other class-level macros
In hanami-controller 1.3.x, we have the following **public class methods** available inside `Hanami::Action` subclasses:

- `before`, `append_before`, `prepend_before`, `after`, `append_after`, `prepend_after` (via the `Action::Callbacks` module)
- `expose` (via `Action::Exposable`)
- `accept` (via `Action::Mime`)
- `use` (via `Action::Rack`)
- `handle_exception` (via `Action::Throwable`)
- `params` (via `Action::Validatable`)

Along with this, every action class has a class-level `configuration` available, which is set by the framework.

As we move to 2.0, we need to consider the overall configuration experience, to ensure it is (a) sensible on its own, (b) provides a reasonable migration path from 1.x, and (c) is coherent with hanami-view as Hanami-controller's counterpart in the overall web stack.

In hanami-controller 2.0, the `configuration` object is largely the same, but there's a different set of public class methods:

- `before`, `append_before`, `prepend_before`, `after`, `append_after`, `prepend_after`
- `params`
- `accept`

This leaves the following 1.x class methods without direct equivalents:

- `expose` — exposures have moved to `Response`, where they can be set with `#[]=`)
- `use` — using rack middleware in actions has been dropped due to unwanted internal complexity (if I understand correctly)
- `handle_exception` — this now exists only on the `configuration` object (which is where it also resides in 1.x, with the class method just delegating to `configuration`)

To address this, I propose we:

- Restore a `handle_exception` class method that delegates to `configuration`
- Move `accept` to `configuration` and rename it `accepted_formats`, then add `accept` back as a class method delegating to `configuration`. This keeps all config together in the same object, and ensures the `accept` values are inheritable, which is not currently the case.
- Leave `params` as a class method (this is key behaviour, and doesn't feel like "config" at all), but update it to be inheritable, so a params defined on a parent class is copied to its children

In doing this, I think we could then outline the following principles for how we manage config and behaviour across hanami-controller and hanami-view

- Most fine-grained configuration lives on class level `config` objects
- Any key or commonly adjusted class-specific behaviour is defined by dedicated class methods. These may be backed by their own state, or may delegate to `config`. Examples of this are the callbacks in `Hanami::Action`, and exposures in `Hanami::View`.
- All class-level behaviour (consisting of both the above) is inheritable.

timriley:
This PR addresses the issues I raised in #389, which was a survey of the class-level interface for `Hanami::Action` and its intersection with `config`.

In #389 I recommended we do the following:

> - Restore a `handle_exception` class method that delegates to `configuration`
> - Move `accept` to `configuration` and rename it `accepted_formats`, then add `accept` back as a class method delegating to `configuration`. This keeps all config together in the same object, and ensures the `accept` values are inheritable, which is not currently the case.
> - Leave `params` as a class method (this is key behaviour, and doesn't feel like "config" at all), but update it to be inheritable, so a params defined on a parent class is copied to its children

Let's look at each of these now:

> - Restore a `handle_exception` class method that delegates to `configuration`

I've done this here in 88277305fb67d6e3a308b47648cd2d42d4f4d56b.

> - Move `accept` to `configuration` and rename it `accepted_formats`, then add `accept` back as a class method delegating to `configuration`. This keeps all config together in the same object, and ensures the `accept` values are inheritable, which is not currently the case.

This was done in #392.

> - Leave `params` as a class method (this is key behaviour, and doesn't feel like "config" at all), but update it to be inheritable, so a params defined on a parent class is copied to its children

I've done this here in 42ea7c8a3c61bd6578dfadcf51d2dd1397f6f440.

With all of these completed, we now have:

- Class methods for all the most frequently used config-like directives (`accept`, `handle_exceptions`) and other behavioural concerns (`params`)
- As well as consistent inheritance for all of these: they will all be copied into any subclasses

This should sort out any awkwardness people have felt in setting up their action classes!

Resolves #389.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -287,6 +287,14 @@ module Hanami
       config.accepted_formats = formats
     end
 
+    # @see Config#handle_exception
+    #
+    # @since 2.0.0
+    # @api public
+    def self.handle_exception(...)
+      config.handle_exception(...)
+    end
+
     # Returns a new action
     #
     # @since 2.0.0
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -266,7 +266,7 @@ module Hanami
     #
     # @since 0.1.0
     #
-    # @see Hanami::Action::Configuration#format
+    # @see Config#format
     #
     # @example
     #   require "hanami/controller"
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -99,6 +99,10 @@ module Hanami
           include Validatable if defined?(Validatable)
         end
       end
+
+      if instance_variable_defined?(:@params_class)
+        subclass.instance_variable_set(:@params_class, @params_class)
+      end
     end
 
     # Returns the class which defines the params
@@ -112,7 +116,7 @@ module Hanami
     # @api private
     # @since 0.7.0
     def self.params_class
-      @params_class ||= BaseParams
+      @params_class || BaseParams
     end
 
     # Placeholder implementation for params class method
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -284,7 +284,7 @@ module Hanami
     #   # When called with "application/json" => 200
     #   # When called with "application/xml"  => 415
     def self.accept(*formats)
-      config.accepted_formats = *formats
+      config.accepted_formats = formats
     end
 
     # Returns a new action
--- a/spec/unit/hanami/action/params_spec.rb
+++ b/spec/unit/hanami/action/params_spec.rb
@@ -499,7 +499,7 @@ RSpec.describe Hanami::Action::Params do
     end
   end
 
-  describe "inhertiance" do
+  describe "inheritance" do
     let(:action) { Class.new(WhitelistedParamsAction).new }
 
     it "uses the params defined in the parent class" do
--- a/spec/unit/hanami/action/params_spec.rb
+++ b/spec/unit/hanami/action/params_spec.rb
@@ -498,4 +498,13 @@ RSpec.describe Hanami::Action::Params do
       expect { params.errors.add(:book, :code, "is invalid") }.to raise_error(ArgumentError, %(Can't add :book, :code, "is invalid" to {:book=>["is missing"]}))
     end
   end
+
+  describe "inhertiance" do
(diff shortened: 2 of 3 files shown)


timriley:
[Check out the deal with `Action.params`](https://trello.com/c/xvCYJUMi/174-check-out-the-deal-with-actionparams)

lib/hanami/action.rb
```diff
@@ -284,7 +288,15 @@ def self.prepend_after(...)
     #   # When called with "application/json" => 200
     #   # When called with "application/xml"  => 415
     def self.accept(*formats)
-      config.accepted_formats = *formats
+      config.accepted_formats = formats
+    end
```
solnic, line 292: Should this really override something that was possibly already set?

spec/unit/hanami/action/params_spec.rb
```diff
@@ -498,4 +498,13 @@
       expect { params.errors.add(:book, :code, "is invalid") }.to raise_error(ArgumentError, %(Can't add :book, :code, "is invalid" to {:book=>["is missing"]}))
     end
   end
+
+  describe "inhertiance" do
```
solnic, line 502:
```suggestion
  describe "inheritance" do
```

lib/hanami/action.rb
timriley: @solnic that's a good question!

Seems like this matches the behaviour from 1.3.x (and I presume all the prior versions), in that calling `accept` ([see the 1.3.x implementation here](https://github.com/hanami/controller/blob/2496ac797237d7320b96048a3c9e3fbebb00532c/lib/hanami/action/mime.rb#L166)) will replace any other previous values ([see where that happens in 1.3.x](https://github.com/hanami/controller/blob/2496ac797237d7320b96048a3c9e3fbebb00532c/lib/hanami/controller/configuration.rb#L446-L448)).

Given this is pre-established behaviour, I'm inclined to leave it as it is, at least for the sake of this PR.

timriley: I'll merge this in for now, since it's not changing previous behaviour. We can still adjust this if we like through a future PR.

Issue #257, opened 2018-04-16
Rename `Hanami::Action.content_type` into `.accept`


jodosha:
With https://github.com/hanami/controller/pull/247 (v1.2.0) we introduced `Hanami::Action.content_type` which is a DSL that restrict the access based on request's `Content-Type` of the payload and halt with `415 Unsupported Media Type` in case of mismatch.

On the other hand we already had `.accept` (v0.1.0), which was built to do the exact same thing, but it targets `Accept` header instead.

To recap: the actual `Hanami::Action.content_type` has the **right** implementation, but it should be named `.accept` instead.

```ruby
class MyAction < Hanami::Action
  accept :json

  def call(*)
    # ...
  end
end
```

This will let to pass only the requests that have a matching `Content-Type` header.

---

This change **MUST** go to `unstable` branch.

---

Ref https://github.com/hanami/controller/pull/247
Ref https://github.com/hanami/controller/commit/fd9e0808ff5fe6798836d9e9b7cab20267383b18

timriley:
Thanks @tak1n, closing this now :)

PR #396, opened 2022-10-17
Reenable MIME specs and check both `Accept` and `Content-Type` headers checks via `accept`

timriley:
## Background

The way our `accept` macro has worked has taken over time has taken a few different shapes.

- In 1.3.x it worked on the `Accept` header only ([see here for code](https://github.com/hanami/controller/blob/2496ac797237d7320b96048a3c9e3fbebb00532c/lib/hanami/action/mime.rb#L168-L172)) and returned a 406 response for non-matching types.
- However, 1.3.x _also_ had a `content_type` macro that operated on the `Content-Type` header ([see code](https://github.com/hanami/controller/blob/2496ac797237d7320b96048a3c9e3fbebb00532c/lib/hanami/action/mime.rb#L198-L204)) and returned a 415 response for non-matching types.
- In the very earliest phase of 2.0 development, when @jodosha was completely overhauling the shape of hanami-controller, we removed the `content_type` macro and left the `accept` macro only (in [this commit](https://github.com/hanami/controller/commit/85e927a694998e5ebfcbad216d7bdec4109b6806), which is rather large). The `accept` macro and its behaviour were left largely unchanged at this point.
- Then not too long after, we changed the `accept` macro to no longer operate on the `Accept` header and instead look at `Content-Type` only, and return a 415 response for all non-matching types (see [this PR](https://github.com/hanami/controller/pull/265), addressing [this issue](https://github.com/hanami/controller/issues/257))

## Shortcomings 

I ran into some trouble with this early 2.0-phase implementation, because it meant that if you put `accept :json` into a base action class (e.g. for every action class within an API service to inherit from), then it would reject _any_ request without a `Content-Type: application/json` header.

Requiring a `Content-Type` header like this would make sense for POST, PUT, PATCH requests, that is, any kind of request that contains a request body and needs to describe its content type.

However, `GET` and other requests without bodies are likely _not_ to have `Content-Type` headers. These requests would more likely have an `Accept` header only.

## Changes

I believe that when an action contains `accept :json`, then it should be able to successfully handle a request with only an `Accept: application/json` header, and no `Content-Type`. It should also handle weighted `Accept` headers too, like `text/html, application/xhtml+xml, application/xml;q=0.9`.

This PR changes the way we enforce the accepted formats configured with `accept`. It:

- Checks the `Accept` header first. If it is present, it will check if any any of the header's MIME types are acceptable per the configured `accepted_formats`. If none are acceptable, then the request will be a 406 error.
- If the `Accept` is not present, it then checks `Content-Type`. If this header is present, but checks to see if the Content-Type MIME type is accepted able per the configured `accepted_formats`. If it is not acceptable, then the request will be a 415 error.
  - If the `Content-Type` header is not present, then it will substitute this with the MIME type for the `default_request_format`, if configured.
- Lastly, if there is no `Accept`, no `Content-Type`, no `default_request_format`, then the request is accepted regardless, with the idea here being that the client has no care or opinion about the format they want, and our Action class can return whatever content it likes in its response.

I believe this strikes a balance that should satisfy the spread of real-world requests. It captures the spirit of _both_ the original `accept` and `content_type` macros while ensuring we still keep the slimmer API surface area of the `accept` macro only.

Certainly if this were implemented earlier, I could have avoided hacking in a workaround inside the app in which I discovered this issue.

## Implementation

This PR has a few key commits that are worth reviewing:

- 17ee819 restores all of our mime_spec.rb examples, most of which were previously disabled. I made no changes to these specs apart from changing their expected response code from 406 to 415 (which shows how they must have been targeting the 1.3.x behaviour still). Interestingly, I also to had to our code to check the `Accept` header too, for all specs to pass.
- d74b4fa implements the new algorithm I described above, and updates the specs to match this behaviour.

Other commits here are mostly about tidying up what has felt like a bit of an unloved area from our 2.0 development phase.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -417,7 +417,7 @@ module Hanami
     # @since 2.0.0
     # @api private
     def enforce_accepted_mime_types(request)
-      return if config.accepted_formats.none?
+      return if config.accepted_formats.empty?
 
       Mime.enforce_accept(request, config) { return halt 406 }
       Mime.enforce_content_type(request, config) { return halt 415 }
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -417,9 +417,10 @@ module Hanami
     # @since 2.0.0
     # @api private
     def enforce_accepted_mime_types(request)
-      return unless config.accepted_formats.any?
+      return if config.accepted_formats.none?
 
-      Mime.accepted_mime_type?(request, config.accepted_mime_types, config) or halt 415
+      Mime.enforce_accept(request, config) { return halt 406 }
+      Mime.enforce_content_type(request, config) { return halt 415 }
     end
 
     # @since 2.0.0
--- a/lib/hanami/action/constants.rb
+++ b/lib/hanami/action/constants.rb
@@ -114,13 +114,13 @@ module Hanami
     #
     # @since 0.1.0
     # @api private
-    HTTP_ACCEPT          = "HTTP_ACCEPT"
+    HTTP_ACCEPT = "HTTP_ACCEPT"
 
     # The header key to set the mime type of the response
     #
     # @since 0.1.0
     # @api private
-    CONTENT_TYPE         = ::Rack::CONTENT_TYPE
+    CONTENT_TYPE = ::Rack::CONTENT_TYPE
 
     # The default mime type for an incoming HTTP request
     #
@@ -152,7 +152,7 @@ module Hanami
     #
     # @since 2.0.0
     # @api private
-    ETAG          = ::Rack::ETAG
+    ETAG = ::Rack::ETAG
 
     # @since 2.0.0
     # @api private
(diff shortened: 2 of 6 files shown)


timriley:
[`accept :json` should be forgiving on requests that have no request bodies (or offer more expansive behavior, encompassing all request methods)](https://trello.com/c/KujFWMWw/113-accept-json-should-be-forgiving-on-requests-that-have-no-request-bodies-or-offer-more-expansive-behavior-encompassing-all-reques)

lib/hanami/action.rb
```diff
@@ -417,9 +417,10 @@ def _requires_no_body?(res)
     # @since 2.0.0
     # @api private
     def enforce_accepted_mime_types(request)
-      return unless config.accepted_formats.any?
+      return if config.accepted_formats.none?
```
solnic, line 420:
```suggestion
      return if config.accepted_formats.empty?
```

this is faster

solnic (approved):

lib/hanami/action.rb
timriley: Thanks, @solnic! Fixed, and I'll remember this for next time :)

PR #397, opened 2022-10-21
Fix ci slack notifications (only on main, not all branches)

cllns:
I haven't tested this since it's a CI thing, but I think it'll work, and will confirm after this is merged.

(The issue here is that in our private Hanami Slack, our #ci issue is reporting failed builds for all branches, which adds a lot of noise and we only want the notifications if `main` is broken.

I also realized the message says ` <|View Workflow>` but without a working link. Turns out we need to specify the token to do that. I trust that GitHub filters it out properly so it won't be public.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Run all tests
         run: script/ci
       - uses: ravsamhq/notify-slack-action@v1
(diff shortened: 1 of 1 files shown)


timriley:
This'll be blessed relief, thank you for looking into this @cllns! Will merge this in and just see how this goes.

PR #380, opened 2022-05-18
Add read only flash & session on request object

tak1n:
Branched off from the work at https://github.com/hanami/controller/pull/379

PR #398, opened 2022-10-21
Add read only flash & session on request object (updated for latest main)

timriley:
Taking #380 and rebasing over main so we can review here.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -85,15 +85,6 @@ module Hanami
 
     # @!scope instance
 
-    # Flash session key
-    #
-    # @since 2.0.0
-    # @api private
-    #
-    # @see Hanami::Action::Request#flash
-    # @see Hanami::Action::Response#flash
-    FLASH_SESSION_KEY = :_flash
-
     # Override Ruby's hook for modules.
     # It includes basic Hanami::Action modules to the given class.
     #
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -85,6 +85,15 @@ module Hanami
 
     # @!scope instance
 
+    # Flash session key
+    #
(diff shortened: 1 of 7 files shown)


timriley:
My plan for this is to have it only freeze the flash object on the request, and leave the session alone.

I've checked and `Request#session` and `Response#session` _always_ return the same object, so even though it is unintuitive, it is still _safe_ to modify the session from the request object. And by not attempting to freeze it and turn it into a read-only structure on the request, we avoid e.g. tampering with key types like this PR did previously (i.e. symboling keys), which means the user can depend on their session keys being exactly how they left them.

It's different for flash, however. `Request#flash` and `Response#flash` will always return flash objects with different `@next` ivars, so it is unsafe to set a flash value on the request; that value will _not_ be persisted to the next request.

We could deal with this in a couple of ways:

- Freeze the flash on the `Request` object, which is how this PR does it currently
- Use a different Flash::ReadOnly style of object on `Request`, which doesn't have any of the methods defined to mutate it
- Initialise the flash object _outside_ the request/response objects, and pass it in as a value when they're being initialized, ensuring they're both using the same object, which would mean that—just like session—we don't end up doing anything special on the request side, and if the flash was mutated there for any reason, we know it'll be properly persisted to the session for the next request

solnic:
Last option seems like the best one. Consistent with how session works and less objects to allocate.

timriley:
@solnic Thanks for your feedback! I agree, and I've gone with that approach (though taking a slightly different implementation path) in #399. Please check it out!

timriley:
Closing in favour of #399.

timriley:
@tak1n Thank you so much for your work here! It was just what we needed to come up with a good working approach for 2.0; see #399 for how we'll handle this :)

PR #399, opened 2022-10-22
Delegate to Request's flash and session from Response

timriley:
This PR addresses some inconsistencies we had in how Request and Response accessed session features, namely the `session` itself, as well as `flash`:

- Until now, `session` has been accessible on both Request and Response,
- But `flash` was available on on the Response only (which was problematic, because Hanami users may want to see the previously-set flash state by interacting with the request)

Earlier I had an idea to these objects available consistently on both Request and Response, but make the Request's version of these objects read-only (see [this Trello card](https://trello.com/c/LSi57W0a/81-add-read-only-flash-and-session-objects-to-request-rationalize-request-response-overall)), which @tak1n implemented in #380. I attempted to clean this up and prepare it for merge in #398, but in doing so, I realised we'd be better served by simply ensuring the request and response operated on the _same_ objects at all times, ensuring that all changes would always be properly persisted through to the next request, regardless of from where they were made.

This PR makes this change. Now, both `session` and `flash` are canonically defined on Request (which is consistent with how `Rack::Request` is modelled — it's the owner of the `#session` method). These same methods are also defined on `Response`, but they simply delegate to the request object, which Response has access to.

This is a much simpler and easier to maintain implementation—and only a very small change from where we are now—so I'd be very confident for us to slip in right before beta4.

While I'm here, I've also removed `Response#request_id`, since we're not using it anywhere, it didn't exist previously in 1.3.x, and `Response` already has access to the `request` anyway, if really needed.

**For reviewers:** this PR has a clean commit history, with a couple of tidy-ups along the way made in separate commits.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -319,8 +319,12 @@ module Hanami
 
       halted = catch :halt do
         params   = self.class.params_class.new(env)
-        request  = build_request(env: env, params: params, sessions_enabled: sessions_enabled?)
-        response = build_response(
+        request  = Request.new(
+          env: env,
+          params: params,
+          sessions_enabled: sessions_enabled?
+        )
+        response = Response.new(
           request: request,
           config: config,
           content_type: Mime.calculate_content_type_with_charset(config, request, config.accepted_mime_types),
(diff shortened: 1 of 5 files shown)


timriley:
[Add read-only flash and session objects to Request, rationalize Request/Response overall](https://trello.com/c/LSi57W0a/81-add-read-only-flash-and-session-objects-to-request-rationalize-request-response-overall)

solnic (approved):
🚀

timriley:
Rebased to handle a merge conflict. Thanks for the review @solnic! I'm going to bring this in now.

PR #401, opened 2022-11-03
Update API docs for 2.0.0

timriley:
Ensure we have 100% coverage for API docs in hanami-controller (and that methods have been checked for appropriate `@api` public or private tags).

This PR does include a couple of small code changes:

In [eea8ef5](https://github.com/hanami/controller/pull/401/commits/eea8ef59ebc1c82c1b20f1f7e4921edd630b3efd), it rearranges the errors into `hanami/action/errors.rb`. `Hanami::Controller::UnknownFormatError` becomes `Hanami::Action::UnknownFormatError` (other error names remain unchanged).

In [6686ec2](https://github.com/hanami/controller/pull/401/commits/6686ec26c8d7d40709bcd2b98b2d8c0680b84006), we remove a CONTENT_TYPE constant that we don't really need internally, because we can use Rack's own `#content_type` method instead of manually referring to the header.

In [ca3761c](https://github.com/hanami/controller/pull/401/commits/ca3761cb54cfb1bbca6589ee8ff9efc45dc40921), it changes the way we assign the response format via `Response#format.

Previously this felt a bit disjointed. You had to use an `Action#format` helper inside your `#handle` method and pass its values to the response:

```ruby
response.format = format(:json)
```

This is a detail that I expect many people will overlook. It's also mostly about associating a format to a content type, which we already have enough information to do inside the response. So now you can just do this:

```ruby
response.format = :json
```

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,5 +1,2 @@
---protected
--
-README.md
-LICENSE.md
-lib/**/*.rb
+--markup=markdown
+--plugin junk
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,9 @@ source "http://rubygems.org"
 gemspec
 
 unless ENV["CI"]
-  gem "byebug", require: false, platforms: :mri
-  gem "yard",   require: false
+  gem "byebug", platforms: :mri
+  gem "yard"
+  gem "yard-junk"
 end
 
 gem "hanami-utils",  github: "hanami/utils", branch: "main"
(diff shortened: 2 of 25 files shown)


timriley:
[Prepare API docs for hanami-controller](https://trello.com/c/uoKEs5ez/264-prepare-api-docs-for-hanami-controller)

jodosha (approved):
@timriley LTGM 👍

PR #402, opened 2022-11-13
Convert to Zeitwerk for code loading

timriley:
Covert the gem to use Zeitwerk for code loading. To do this:

- Add zeitwerk to the gemspec
- Add `lib/hanami-controller.rb` so the gem follows a conventional structure
- Set up the Zeitwerk loader inside `lib/hanami/action.rb` (call it `.gem_loader` instead of `.loader` like we have done in other gems, since `Hanami::Action` is a class with other responsibilities, and `loader` is too generic a term)
- Move conditional requires below the `gem_loader.setup` line to ensure any internal autoloader work
- Remove all now-unneeded manual requires for files inside the gem

This is not a breaking change because any existing `require "hanami/action"` will continue to work as expected, since that is where the Zeitwerk loader setup is done.

Having the loader created there also means we can trivially rename this gem to hanami-action at some point in the future, since there are now basically 0 files inside hanami/controller/ (just the version.rb, nothing else).


timriley:
[Zeitwerkify hanami-controller](https://trello.com/c/LeEUnLjL/328-zeitwerkify-hanami-controller)

PR #403, opened 2022-11-14
Add single `formats` setting to replace `accepted_formats`, `default_request_format`, `default_response_format`

timriley:
UPDATE: Thanks to @jodosha's suggestion, we've built further on the ideas below and encapsulated all format configuration under a new `Config::Formats` instance returned from `config.formats`. It has its own API and it's very nice. We've kept the `config.format(...)` method as a short cut to assign `config.formats.values`.

---

So this PR started with my adding `Config#use_formats` as a "macro" to configure `accepted_formats`, `default_request_format` and `default_response_format` all together, since that is what typical well-behaved actions would need if they are going to be format-aware.

This led to a bit of a naming conversation: `#use_formats` felt awkward to me, so @jodosha suggested we rename the existing `Config#format` to `#add_format`, which would then allow us to name this new macro `#format`, which is clearly the best name for it.

@jodosha also hinted that it might make more sense to turn this macro into a real setting, too, since the trouble with a macro without any concrete underlying variable is that it becomes hard to authoritatively read a value back from it later.

This all sounded great to me, so I took a pass at it, and it turns out this has unlocked the chance to greatly simplify our actions config. It was very much a "one thing led to another" kind of scenario, so let me walk you through the steps (and commits):

1. 0ed04f3505c0974537c850e8d67b603703764b92 — We start with the `Config#format` -> `#add_format` rename. At the same time, let's rename the underlying `formats` setting to `format_mappings` since we didn't want any confusion with this new setting we'll be introducing later (plus, `format_mappings` is a much more self-descriptive name for this settings' purpose).
2. b00ee8f95e28b147d80b93bea0f89f3bdf899530 — With that done, we can rename `Config#use_formats` to `Config#format`, which was our agreed ideal name for this. Also add `Action.format` as a class-level shortcut. However, at this point, this method is still a "macro" and not a simple setter method backed by a concrete setting, so let's start by seeing if we can reduce what this macro is doing.
3. 6df1e391ec4a8566b3c44276441151e55e9c9012 — To begin with, we can entirely remove the `default_request_format` setting. This turned out to be entirely useless: it was only _really_ used inside `Mime.enforce_content_type`, but it turns out the `if content_type.nil?` guard statement below is already doing enough for us here. We don't need this setting.
4. 5fae6fcc25f06b6e66ba7b28ecc52d95a5111e7d — From here, we can make a proper `formats` setting to hold the formats we're passing to `Config#format`. We can also remove `default_response_format` and instead use the first entry from `formats` to take its place.
5. a23cb9330bcb5d3d2f96b90c066725238af48bbd — As a last step, we can now remove `Action.accept` and the `accepted_formats` setting, since we already have a list of formats we can use for this: the new `formats` setting! 
 
So by the time we get to the end of this, the new `format` method (and the `formats` setting it populates) are just an evolution of what we had with `accept`/`accepted_formats` previously, with the added benefit of setting the default response format at the same time (which can easily be overridden per action if ever needed). And the removal of all the other settings makes for a much tighter overall experience and a much easier to understand internal implementation.

I'm honestly super jazzed that we got to this point! I think it's going to make actions so much more streamlined and usable for devs working with Hanami 2.0!

Let me just re-state where we were before:

```ruby
  config.accepted_formats = [:json] # or accept(:json)
  config.default_request_format = :json
  config.default_response_format = :json
end
```

Then with my initial pass at this, we got to:

```ruby
  config.use_formats :json
end
```

And now we're at:

```ruby
  format :json
end
```

And with the added benefit of `accept`, `default_request_format` and `default_response_format` **all being fully removed**, so that's three whole different settings and concepts our users no longer have to try and grapple with. Those three settings were added at different points of the Hanami 1.x lifecycle to address various needs/bugs that arose, but with 2.0 we have the chance to clear the slate and here we've arrived at a much more cohesive outcome: a solitary setting that we can have impact all aspects of the action request handling as required, and just a single thing our users have to learn about.

Pls let me know what you think 😄 

---

_Original description text follows:_

Add a `Config#use_formats` "macro" to make it easy to configure several settings that are usually done together:

- `accepted_formats` (which can also be set via `Action.accept`)
- `default_request_format` (which IMO probably doesn't even need to _exist_ as a setting, but it's too late for that now, we can reconsider it later)
- `default_response_format`

So instead of doing this:

```ruby
config.accepted_formats = [:json]
config.default_request_format = :json
config.default_response_format = :json
```

You can just do this:

```ruby
config.use_formats :json
```

We also support passing multiple formats:

```ruby
config.use_formats :json, :html
```

In this case, the first format is the default format, and will be the format assigned as the default request/response format. The full array of formats is passed to `accepted_formats`.

In this present state, this PR presents a pure _addition_ only: `#use_format` is a "macro" that serves as convenience and reduction in boilerplate for common configuration requirements.

An open question here is what we should name this. I'm not completely happy with `config.use_formats`. What else could we use?

- `config.action_format :json` (not bad, I actually don't mind it, I might be inclined to move to this unless someone presents a better alternative — it's a good in that it hints at the idea of "everything about this action is oriented around <this format>")
- `config.format! :json` (nah, the meaning of the `!` is ambiguous)
- _others?_

One other name I toyed with earlier was `accept_formats`. This was nice in that there was a clear relationship to one of the settings that it assigned (`accepted_formats`). I even went so far as to change `Action.accept` to call this new method, but that did break 2 specs (I suspect the change in behaviour would be tolerable, but we should be extra wary of such things at this point). The downside to `accept_formats` as a name, however, is that it doesn't quite encompass the idea that _other_ aspects aside from `accepted_formats` are also being configured at the same time.

Another thing I considered here was whether this method should also accomodate the registration of custom MIME/content type strings with action names, but I felt that would be putting way too much responsibility into one place, and it'd complicate the call signature. I think we're better off leaving the registration of format/mime type pairings to the dedicated `Config#format` method.

So a couple questions to leave with you:

1. Are you happy for the idea behind this method? I think it will be important for streamlining the experience of people building APIs with Hanami 2, which is going to be our major use case for the first release.
10. Do you have any opinions on naming? Absent any strong opinions I think I'll rename it to `action_format`.

Once we're happy I'll tweak the name, tidy the docs, and merge.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -58,7 +58,7 @@ module Hanami
 
     # See {Config} for individual setting accessor API docs
     setting :handled_exceptions, default: {}
-    setting :formats, default: Config::Formats.new, cloneable: true
+    setting :formats, default: Config::Formats.new, mutable: true
     setting :default_charset
     setting :default_headers, default: {}, constructor: -> (headers) { headers.compact }
     setting :cookies, default: {}, constructor: -> (cookie_options) {
@@ -70,8 +70,8 @@ module Hanami
       Pathname(File.expand_path(dir || Dir.pwd)).realpath
     }
     setting :public_directory, default: Config::DEFAULT_PUBLIC_DIRECTORY
-    setting :before_callbacks, default: Utils::Callbacks::Chain.new, cloneable: true
-    setting :after_callbacks, default: Utils::Callbacks::Chain.new, cloneable: true
+    setting :before_callbacks, default: Utils::Callbacks::Chain.new, mutable: true
+    setting :after_callbacks, default: Utils::Callbacks::Chain.new, mutable: true
 
     # @!scope class
 
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -311,7 +311,7 @@ module Hanami
         response = build_response(
           request: request,
           config: config,
-          content_type: Mime.calculate_content_type_with_charset(config, request, config.formats.accepted_mime_types),
+          content_type: Mime.response_content_type_with_charset(request, config),
           env: env,
           headers: config.default_headers,
           sessions_enabled: sessions_enabled?
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -311,7 +311,7 @@ module Hanami
         response = build_response(
           request: request,
           config: config,
-          content_type: Mime.calculate_content_type_with_charset(config, request, config.accepted_mime_types),
+          content_type: Mime.calculate_content_type_with_charset(config, request, config.formats.accepted_mime_types),
           env: env,
           headers: config.default_headers,
           sessions_enabled: sessions_enabled?
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -59,9 +59,6 @@ module Hanami
     # See {Config} for individual setting accessor API docs
     setting :handled_exceptions, default: {}
     setting :format_mappings, default: Config::DEFAULT_FORMAT_MAPPINGS
-    setting :default_request_format, constructor: -> (format) {
-      Utils::Kernel.Symbol(format) unless format.nil?
-    }
     setting :default_response_format, constructor: -> (format) {
       Utils::Kernel.Symbol(format) unless format.nil?
     }
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -278,6 +278,14 @@ module Hanami
       config.after_callbacks.prepend(...)
     end
 
+    # TODO: document
+    def self.format(...)
+      config.format(...)
+    end
+    # class << self
+    #   alias_method :accept, :format
+    # end
+
     # Restrict the access to the specified mime type symbols.
     #
     # @param formats[Array<Symbol>] one or more symbols representing mime type(s)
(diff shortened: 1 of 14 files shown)


timriley:
[Consider a convenience method to make controller setup for the case of accepting/responding with a single format](https://trello.com/c/lWNdGriJ/334-consider-a-convenience-method-to-make-controller-setup-for-the-case-of-accepting-responding-with-a-single-format)

spec/support/fixtures.rb
```diff
@@ -1610,20 +1610,8 @@ def handle(_req, res)
...
-    def handle(*, res)
-      res.format = :xml
-    end
-
-    private
-
-    def default_response_format
-      :json
-    end
-  end
```
timriley, line 1623: The test for this class made no sense anymore, so I removed them both.

spec/support/fixtures.rb
```diff
@@ -1585,7 +1585,7 @@ def handle(req, res)
   end
 
   class CustomFromAccept < Hanami::Action
-    config.add_format custom: "application/custom"
```
jodosha: @timriley If you don't like this removal, we can still reintroduce `#add_format`.

lib/hanami/action.rb
```diff
@@ -58,8 +58,7 @@ def self.gem_loader
 
     # See {Config} for individual setting accessor API docs
     setting :handled_exceptions, default: {}
-    setting :format_mappings, default: Config::DEFAULT_FORMAT_MAPPINGS
```
jodosha: @timriley The main point of this commit is to have only `config.formats` instead of two: `format_mappings` and `formats`.

PR #411, opened 2023-01-02
Fix for accepting all content types

pat:
I'm in the process of migrating an API within a Hanami app from Roda to hanami-router/hanami-controller. And while I do prefer to be strict about the content received by the API, I also don't entirely trust the third-parties using it will send through correct content types - and I don't want to be breaking their access via their current behaviour.

So, I've set `format :all` in the relevant action superclass… and my tests break. 😭

From what I can see: when we use `:all` as a format, we're comparing against the mime-type strings of `*/*` and `application/octet-stream`. rack-test sends through a default content type of `application/x-www-form-urlencoded`. But it doesn't get matched against the catch-all `*/*`, because [the order of arguments to `Rack::Mime.match?`](https://github.com/rack/rack/blob/v3.0.3/lib/rack/mime.rb#L23-L35) actually matters: if you want the wildcard behaviour to apply, the wildcards need to be in the _second_ argument.

Now that I've flipped the arguments around, everything's happy.

All that said, I'm wading into unfamiliar code, so: have I misunderstood things? Should I be testing it in a different way? Should the other uses of `Rack::Mime.match?` be modified as well? (I feel like they should, but unsure how to write a breaking test to then fix with the switch of arguments)

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -237,7 +237,7 @@ module Hanami
         # @api private
         def accepted_mime_type?(mime_type, config)
           accepted_mime_types(config).any? { |accepted_mime_type|
-            ::Rack::Mime.match?(accepted_mime_type, mime_type)
+            ::Rack::Mime.match?(mime_type, accepted_mime_type)
           }
         end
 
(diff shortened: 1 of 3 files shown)


jodosha (approved):
@pat This is a good catch. I double-checked the `Rack::Mime.match?` signature, and yes, we had flipped arguments.

jodosha:
@pat 
> I'm in the process of migrating an API within a Hanami app from Roda to hanami-router/hanami-controller.

❤️ Thank you.

pat:
Glad to know I got it right 😅 and thank you for merging :) (and all the Hanami work! 👏🏻)

PR #412, opened 2023-01-04
Support pattern-matching Params

alassek:
I think this is a pretty straightforward feature enhancement, currently you have to call `#to_h` on it to destructure keys, which feels like unnecessary friction.

### DIFF
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -243,6 +243,15 @@ module Hanami
         @params
       end
       alias_method :to_hash, :to_h
+
+      # Pattern-matching support
+      #
+      # @return [::Hash]
+      #
+      # @since 2.1.0
+      def deconstruct_keys(_keys)
+        @params
+      end
     end
   end
 end
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -248,9 +248,9 @@ module Hanami
       #
       # @return [::Hash]
       #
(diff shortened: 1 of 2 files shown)


lib/hanami/action/params.rb
```diff
@@ -243,6 +243,15 @@ def to_h
         @params
       end
       alias_method :to_hash, :to_h
+
+      # Pattern-matching support
+      #
+      # @return [::Hash]
+      #
+      # @since 2.1.0
+      def deconstruct_keys(_keys)
```
jodosha, line 252:
```suggestion
      def deconstruct_keys(*)
```

lib/hanami/action/params.rb
```diff
@@ -243,6 +243,15 @@ def to_h
         @params
       end
       alias_method :to_hash, :to_h
+
+      # Pattern-matching support
+      #
+      # @return [::Hash]
+      #
+      # @since 2.1.0
```
jodosha, line 251:
```suggestion
      # @since 2.0.2
```

lib/hanami/action/params.rb
```diff
@@ -243,6 +243,15 @@ def to_h
...
+      def deconstruct_keys(_keys)
+        @params
```
jodosha, line 253:
```suggestion
        to_hash
```

jodosha (approved):
@alassek Excellent enhancement. 👏 Thank you!

PR #416, opened 2023-01-30
Fix error message for missing format (aka MIME Type)

jodosha:
In a Hanami 2.0 app, I add to an action `format :graphql`.

By restarting the app, I got the following error message:

```shell
Cannot find a corresponding MIME type for format :graphql. Configure one via `config.formats.add(graphql: \"MIME_TYPE_HERE\")`.
```

But there are two problems:

  1. It's `config.actions.formats`, not `config.formats`
  2. The `#add` method requires two arguments (`Symbol` and `String | Array<String>`, not one (`Hash`).

### DIFF
--- a/lib/hanami/action/errors.rb
+++ b/lib/hanami/action/errors.rb
@@ -37,7 +37,7 @@ module Hanami
       def initialize(format)
         msg =
           if format.to_s != "" # rubocop:disable Style/NegatedIfElseCondition
-            "Cannot find a corresponding MIME type for format #{format.inspect}. Configure one via `config.formats.add(#{format}: \"MIME_TYPE_HERE\")`." # rubocop:disable Layout/LineLength
+            "Cannot find a corresponding MIME type for format #{format.inspect}. Configure one via `config.actions.formats.add(:#{format}, \"MIME_TYPE_HERE\")`." # rubocop:disable Layout/LineLength
           else
             "Cannot find a corresponding MIME type for `nil` format."
           end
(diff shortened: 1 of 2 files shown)


timriley (commented):
Thanks for finding and fixing this, @jodosha!

There’s actually two issues here:

(1) The incorrect params signature for `#add` is an unequivocal bug, which you’ve properly fixed here.

(2) Whether to show `config.formats` or `config.actions.formats` is another matter. I originally wrote `config.formats` intentionally, because that is the correct thing to write when configuring a `Hanami::Action` class directly:

```ruby
  config.formats.add(:unknown, "mime_type_here")
end
```

`config.actions.formats` only works when you're configuring _all_ actions within an app or slice class body.

So I think we probably do want to change this part of the error message, but to something slightly different.

We could e.g. change this to show _both_ `config.formats` as well as `config.actions.formats`?

We could figure out whether a full Hanami app is being used and show `config.actions.formats` _only_ in that case?

This will apply to both action _and_ view error messaging, so it'd be good to establish a pattern here :)

jodosha:
@timriley, What do you think about this comprehensive error message?

```ruby
message = <<~MSG
Cannot find a corresponding MIME type for format #{format.inspect}.

Configure one via `config.actions.formats.add(:#{format}, "MIME_TYPE_HERE")' in `config/app.rb' to share between actions.

Or make it available only in the current action: `config.formats.add(:#{format}, "MIME_TYPE_HERE")'
MSG
```

jodosha:
@timriley I went ahead with this change. Please let me know. 🙂

timriley (approved):

Issue #420, opened 2023-03-20
Support CSRF by HTTP header


masterT:
It would be nice to be able to configure the CSRF validation so it can be read from a particular HTTP header. 

Let me know if this is something that would fit into the project, I would be happy to contribute. 🙂

jodosha:
@masterT What's the use case?

masterT:
I want to protect a web API consumed by AJAX requests by a browser-based client application. I'm using [Axios](https://github.com/axios/axios) to perform the HTTP requests to the API.

At the moment I'm using a [request interceptor](https://github.com/axios/axios/#interceptors) to set the CSRF token in the query parameter because it is too hard to set the token in the body (as the body can be String, FormData, Object, Buffer, etc.). This does not feel like the right approach for AJAX requests. 

I would like to use the solution proposed by OWASP to [set the CSRF token in a custom HTTP header](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#javascript-guidance-for-auto-inclusion-of-csrf-tokens-as-an-ajax-request-header).

andrewcroome:
For what it's worth, I've also needed to support passing the CSRF token via a header in the past, which I did by patching `Hanami::Actions::CSRFProtection` to add a method similar to: 

```ruby
def request_csrf_token(req)
  req.params[CSRF_TOKEN] || req.get_header("HTTP_X_CSRF_TOKEN")
end
```

Laravel supports this [out of the box](https://laravel.com/docs/10.x/csrf#csrf-x-csrf-token), and it looks like Django allows you to configure [which header to use](https://docs.djangoproject.com/en/4.1/howto/csrf/#using-csrf-protection-with-ajax). 

Rails UJS will [set the header](https://github.com/rails/rails/blob/main/actionview/app/assets/javascripts/rails-ujs.js#L58) on requests too, as per the [security guide](https://guides.rubyonrails.org/security.html).

masterT:
Overriding the method `request_csrf_token(req)` seems like a good way to solve the problem, but the method those not exist.


I will craft something using this solution.

timriley:
This is now done via #422 and released as part of Hanami v2.3! 🎉

PR #422, opened 2023-04-01
feat: Support custom CSRF token retrieval.

masterT:
Add the method `request_csrf_token(req)` in `Hanami::Action::CSRFProtection` to encapsulate the request token retrieval so it can be overridden to change the default behaviour.

Example:

```ruby
module Web::Controllers::Books
  class Create < Web::Action
    def handle(*)
      # ...
    end

    private

    def request_csrf_token(req)
      req.get_header('X-CSRF-Token')
    end
  end
end
```

Reference:
- #420

### DIFF
--- a/lib/hanami/action/csrf_protection.rb
+++ b/lib/hanami/action/csrf_protection.rb
@@ -14,7 +14,10 @@ module Hanami
     #
     # It stores a "challenge" token in session. For each "state changing request"
     # (eg. <tt>POST</tt>, <tt>PATCH</tt> etc..), we should send a special param:
-    # <tt>_csrf_token</tt>.
+    # <tt>_csrf_token</tt> which contain the "challenge" token.
+    #
+    # We can specify how to retreive the token from the request, by overriding <tt>#request_csrf_token</tt>.
+    # This is useful if we want to handle XMLHttpRequest (XHR) requests.
     #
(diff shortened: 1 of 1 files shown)


lib/hanami/action/csrf_protection.rb
```diff
@@ -107,6 +125,34 @@ def set_csrf_token(_req, res)
         res.session[CSRF_TOKEN] ||= generate_csrf_token
       end
 
+      # Get CSRF Token from request.
+      #
+      # By default retreives the token from the request param <tt>_csrf_token</tt>.
+      #
+      # Override this method, for custom handling of the request token retrieval.
+      #
+      # @since 2.X.X
```
masterT, line 134: The version should be set.

timriley:
Thanks for putting this together, @masterT! However, rather than leave this as a hook for users to override, I'd prefer we build in support for checking the header for the CSRF token in addition to the request's params.

Would you like to have a go at doing that?

masterT:
Sure, I will craft something.

masterT:
I did implement your suggestion, this is ready for review. 🙂 @timriley

timriley:
Thanks again for this feature, @masterT! Your second round of changes looks perfect, and I'll be very happy for us to release this with Hanami 2.3 next week :)

PR #423, opened 2023-04-22
Add #session_enabled? to Request and Response

timriley:
Add new `Request#session_enabled?` and `Response#session_enabled?` public methods.

This will be useful for external code needing to check whether the sesison is enabled before interacting with it (like Hanami’s form helpers need to do before attempting to fetch a CSRF token from the request, see https://github.com/hanami/hanami/pull/1305 for details).

Along with this, rename the underlying ivar and argument name from sessions_enabled (plural) to session_enabled (singular). This makes it consistent with the singular `Session` module that we use, as well as the singular `#session` that we expose from the request and response.

Given these are new public methods, these should probably go out in the 2.1.0 release of hanami-controller? Right now I have them marked as `@since x.x.x`.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -309,7 +309,7 @@ module Hanami
         request  = build_request(
           env: env,
           params: params,
-          sessions_enabled: sessions_enabled?
+          session_enabled: session_enabled?
         )
         response = build_response(
           request: request,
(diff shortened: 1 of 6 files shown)


timriley:
[Finish form helpers](https://trello.com/c/RYVt348g/4-finish-form-helpers)

PR #428, opened 2023-09-20
`Hanami::Action::Config#root` to match `Hanami.app.root`

jodosha:


jodosha:
Found a cleaner way to achieve this. Thanks @timriley !

PR #429, opened 2023-09-27
`Hanami::Action::Config#root`: don't check realpath existence

jodosha:
`Hanami::Action::Config#root`: don't check `Pathname#realpath` existence to simplify the boot process of Hanami.

Ref https://github.com/hanami/hanami/pull/1333#discussion_r1335857970


timriley (approved):
👍🏼

PR #431, opened 2023-10-06
Ensure Rack compatibility of `Response#send_file`

### LINKED ISSUE #430: Fix Response#send_file implementation

jodosha:
## Enhancements

* Ensure Rack compatibility for `Hanami::Action::Response#send_file` and `#unsafe_send_file`.
* Make it clear that the object used is `Hanami::Action::Rack::File` and not `Rack::File` from `rack` gem.
* Ensure that previously skipped specs are now passing.

---

Closes #430

### DIFF
--- a/lib/hanami/action/response.rb
+++ b/lib/hanami/action/response.rb
@@ -73,8 +73,7 @@ module Hanami
         @length = 0
         @body   = EMPTY_BODY.dup
 
-        # FIXME: there could be a bug that prevents Content-Length to be sent for files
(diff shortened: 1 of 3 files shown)


timriley (approved):
Thanks for sorting this out, @jodosha! I'm glad this turned out to be less of an issue than I originally imagined. On that note, the clearer scoping of `Action::Rack::File` is a huge win for understandability given rack itself exposes that same-named class 😄 

Congrats on getting some tests unskipped too, a win all round!

Issue #421, opened 2023-03-23
Support for hanami-validations/dry-validations `rule(...)`


Drowze:
In `dry-validations` and `hanami-validations` we can define arbitrary _rules_, which are particularly handy when defining a validation that acts on multiple attributes, e.g.:

```ruby
class EventContract < Dry::Validation::Contract
  params do
    required(:start_date).value(:date)
    required(:end_date).value(:date)
  end

  rule(:end_date, :start_date) do
    key.failure('must be after start date') if values[:end_date] < values[:start_date]
  end
end
```

On `hanami-controller`, I see that we have support for a `params` block within actions, so we can wrap our request params within a hanami-validations `params` block and validate accordingly.
...But I don't see any way to use arbitrary _rules_ on hanami-controller. Is it possible on Hanami 2?

I know it was possible in Hanami 1, e.g.:
```ruby
module Api
  class MyAction
    include Api::Action
    params do
      required(:start).filled(:int?)
      required(:end).filled(:int?)

      rule(dates: %i[start_date end_date]) do |_start, _end|
        _end.gt?(value(:_start))
       end
     end

     def call(params)
       # ...
     end
  end
end
```

In Hanami 2 however, I'm currently simply using concrete classes inheriting from Dry::Validations::Contract and manually instantiating and validating the parameters, but I would like to see a more supported way if possible 🤔

timriley:
This was completed in #453 and released as part of Hanami 2.2! 🎉

PR #434, opened 2023-11-16
Allow `Dry::Validation::Contract` rule definitions in `Hanami::Action`

danhealy:
For a `Validatable` `Hanami::Action` (including `hanami-validations`), calls to `rule` inside the `Hanami::Action` are passed through to the `Dry::Validation::Contract` class via the `Hanami::Action::Params` class. 

```ruby
class Event < Hanami::Action
  params do
    required(:start_date).value(:date)
    required(:end_date).value(:date)
  end

  # Rules must be defined after the params schema.
  rule(:start_date, :end_date) do
    if start_date > end_date
      base.failure('event cannot end before it starts')
    end
  end

  def handle(req, *)
    halt 400 unless req.params.valid?
    # ...
  end
end
```

Reversing the order of `params` and `rule` will raise an exception, this is the same constraint that `Dry::Validation::Contract` has.

https://github.com/hanami/controller/issues/421

### DIFF
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -148,23 +148,23 @@ module Hanami
         validations(&blk || -> {})
       end
 
-      # Define full contract validations.  This takes precendence over the older `params` DSL.
+      # Define a validation rule.
       #
-      # @param blk [Proc] the validation definitions
+      # @param blk [Proc] the rule definition
       #
-      # @since 2.1.0
+      # @since 2.x.x
       #
       # @example
       #   class Event < Hanami::Action
-      #     contract do
-      #       params do
-      #         required(:start_date).value(:date)
-      #         required(:end_date).value(:date)
-      #       end
-      #       rule do
-      #         if start_date > end_date
-      #           base.failure('event cannot end before it starts')
-      #         end
+      #     params do
+      #       required(:start_date).value(:date)
+      #       required(:end_date).value(:date)
+      #     end
+      #
+      #     # Rules must be defined after the params schema.
+      #     rule(:start_date, :end_date) do
+      #       if start_date > end_date
+      #         base.failure('event cannot end before it starts')
       #       end
       #     end
       #
@@ -173,18 +173,13 @@ module Hanami
       #       # ...
       #     end
       #   end
-      def self.contract(&blk)
-        @_contract = Dry::Validation::Contract.build(&blk || -> {})
-      end
-
-      # Accessor for the validations contract
-      #
-      # @return [Dry::Validation::Contract, NilClass]
-      # 
-      # @since 2.1.0
-      # @api private
-      def self._contract
-        @_contract
+      def self.rule(*keys, &blk)
+        unless self._validator
+          raise ArgumentError.new(
+            "The params schema must be defined prior to adding rules."
+          )
+        end
+        self._validator.class.rule(*keys, &blk)
       end
 
       # Initialize the params and freeze them.
@@ -198,9 +193,9 @@ module Hanami
       def initialize(env)
         @env = env
         super(_extract_params)
-        validation = self.class._contract ? self.class._contract.call(@input) : validate
+        validation = validate
         @params = validation.to_h
-        @errors = Errors.new(self.class._contract ? validation.errors.to_h : validation.messages)
+        @errors = Errors.new(validation.messages)
         freeze
       end
 
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -173,8 +173,8 @@ module Hanami
       #       # ...
       #     end
       #   end
-      def self.rule(*keys, &blk)
-        self._validator.class.rule(*keys, &blk)
+      def self.rule(...)
+        _validator.class.rule(...)
       end
 
       # Initialize the params and freeze them.
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -148,6 +148,45 @@ module Hanami
         validations(&blk || -> {})
       end
 
+      # Define full contract validations.  This takes precendence over the older `params` DSL.
+      #
+      # @param blk [Proc] the validation definitions
+      #
+      # @since 2.1.0
+      #
+      # @example
+      #   class Event < Hanami::Action
+      #     contract do
+      #       params do
+      #         required(:start_date).value(:date)
+      #         required(:end_date).value(:date)
+      #       end
+      #       rule do
+      #         if start_date > end_date
+      #           base.failure('event cannot end before it starts')
+      #         end
+      #       end
+      #     end
+      #
+      #     def handle(req, *)
+      #       halt 400 unless req.params.valid?
+      #       # ...
+      #     end
+      #   end
+      def self.contract(&blk)
+        @_contract = Dry::Validation::Contract.build(&blk || -> {})
+      end
+
+      # Accessor for the validations contract
+      #
+      # @return [Dry::Validation::Contract, NilClass]
+      # 
+      # @since 2.1.0
+      # @api private
+      def self._contract
+        @_contract
+      end
+
       # Initialize the params and freeze them.
       #
       # @param env [Hash] a Rack env or an hash of params.
@@ -159,9 +198,9 @@ module Hanami
       def initialize(env)
         @env = env
         super(_extract_params)
-        validation = validate
+        validation = self.class._contract ? self.class._contract.call(@input) : validate
         @params = validation.to_h
-        @errors = Errors.new(validation.messages)
+        @errors = Errors.new(self.class._contract ? validation.errors.to_h : validation.messages)
         freeze
       end
 
--- a/lib/hanami/action/params.rb
+++ b/lib/hanami/action/params.rb
@@ -174,11 +174,6 @@ module Hanami
       #     end
       #   end
       def self.rule(*keys, &blk)
-        unless self._validator
-          raise ArgumentError.new(
-            "The params schema must be defined prior to adding rules."
-          )
-        end
         self._validator.class.rule(*keys, &blk)
       end
 
--- a/spec/unit/hanami/action/rules_spec.rb
+++ b/spec/unit/hanami/action/rules_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe Hanami::Action do
     let(:response) { action.call(given_input) }
 
     context "given valid input" do
-      let(:given_input) { { book: { code: "abc" } } }
+      let(:given_input) { {book: {code: "abc"}} }
 
       it "is valid" do
         expect(response[:params].valid?).to eq(true)
@@ -31,7 +31,7 @@ RSpec.describe Hanami::Action do
     end
 
     context "given invalid input" do
-      let(:given_input) { { book: { code: nil } } }
+      let(:given_input) { {book: {code: nil}} }
 
       it "is not valid" do
         expect(response[:params].valid?).to eq(false)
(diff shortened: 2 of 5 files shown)


lib/hanami/action/params.rb
```diff
@@ -148,6 +148,40 @@ def self.params(&blk)
...
+      #       halt 400 unless req.params.valid?
+      #       # ...
+      #     end
+      #   end
+      def self.rule(*keys, &blk)
+        unless self._validator
+          raise ArgumentError.new(
+            "The params schema must be defined prior to adding rules."
+          )
+        end
```
danhealy, line 181: I'm not sure if this guard is necessary, or if the one below is sufficient.

parndt:
@danhealy FYI, looks like `main` is the same as `2.1.x` (0 files changed) so I've updated the PR's base base branch to `main` (which is why there are now 4 merge commits which would disappear if you rebased from `main`). Hope this is helpful!

parndt:
@danhealy This patch works exceedingly well for me! Previously I was using this great code from @katafrakt https://github.com/katafrakt/palaver/blob/d1a31fec7dba6638ec798c419cecf35ce826f28e/lib/palaver/action.rb#L18-L25 but I've been able to remove that and all of my `contract do` wrappers and use what I had written inline instead.

lib/hanami/action/params.rb
parndt: @danhealy in my experimentation this one is not required as I see the error even when I remove it as it comes from `validatable.rb` instead:

image

parndt, line 181:
```suggestion
```

danhealy:
Thank you for the feedback @parndt ! I removed the extra guard and rebased on `main`

parndt:
@danhealy awesome, I pushed a fix for rubocop offenses to your branch so that CI will not fail on that.

parndt:
@jodosha @timriley how do you want to handle this? It would be good to get this in for 2.1 but I'll leave that up to you. Either way we should changelog it?

danhealy:
I'll fix the remaining style issues

parndt (dismissed):
This works really well in my app!

parndt:
> I'll fix the remaining style issues

nice, your fixes mirror exactly how I was going to do it.

parndt:
Woohoo ✅ CI 🎉

timriley:
Thanks so much for this, @danhealy! Got it done by conference end, amazing stuff 😍

I'm just about to enter into various metal tubes for 24h, so I'll look at this once I'm back early next week.

timriley:
@danhealy Thanks again for making this! So it turns out in America you don't have to go through customs before leaving the country, which meant I had a bit of bonus computer time! I won't have time to go through the code in detail yet, but I'll leave one thought here for you consideration:

After seeing this in totality, I wonder if a better API for this would be (building from your example):

```
params do
  required(:start_date).value(:date)
  required(:end_date).value(:date)
end

rules do

@parndt I love your enthusiasm here. 😉 So, _technically_, we've frozen 2.1, and our attention should be on making sure any fixes to existing features are squared away before considering other things.

For this feature, we'll also want to make sure we properly test it and get to a point where we're happy with the API. @jodosha's attention has been on our assets fixes and I don't want to rush him with this one either. ☺️

As for the API,

**Comment by timriley** [comment]
@danhealy Thanks again for making this! So it turns out in America you don't have to go through customs before leaving the country, which means I have a bit of bonus computer time! I won't be able to go through the code in detail yet, but do have one thought here for you consideration:

After seeing this in totality, I wonder if a better API for this would be (building from your example):

```ruby
params do
  required(:start_date).value(:date)
  required(:end_date).value(:date)
end

rules do
  rule(:start_date, :end_date) do
    if start_date > end_date
      base.failure('event cannot end before it starts')
    end
  end
end
```

In this way, we just have two clear, well-contained blocks for action params validation: `rules` and `params`. This means they also mirror each other in terms of having pluralised names that each contain one or more of the items they describe (individual param schema entries and individual rules, respectively).

In such an approach, I think we'd just take the block passed to `rules` and eval that in the context of the contract class. In this way, I think it could make the dry-validation integration simpler and more flexible, since inside `rules` gives you the ability to do whatever else you want inside that validation contract context.

Now, if we did take this approach, it does make me wonder if `contract` might be a better name for this method (despite the positive aspects I noted about `rules` before): it'd make it clearer that it's working with the context of the contract class, and it could also work where it's the _only_ block you use, if that's your preference, e.g.

```ruby
contract do
  params do
    required(:start_date).value(:date)
    required(:end_date).value(:date)
  end

  rule(:start_date, :end_date) do
    if start_date > end_date
      base.failure('event cannot end before it starts')
    end
  end
end
```

Having `contract` eval straight into a contract class might also open up interesting avenues where users want to use different schema types, e.g. `json` from https://dry-rb.org/gems/dry-validation/1.10/schemas/. Though I think we'd want to test what that looks like in practice before deciding if that's something we want to allow or discourage.

As you know from our earlier discussions, @danhealy, I think it's worth us thoroughly exploring all our API design options here. As someone who's been active in this space now, I'm interested in your thoughts on the above!

And thanks again for your code contributions — having these changes be real and concrete and _right here in front of our eyes_ is going to help so much for us to get this much needed feature done :)

In the meanwhile, I'll continue to think on this. Will get back to you again when I'm (a) in Australia, and (b) have any further developed thoughts to add.

--

@parndt — I love your enthusiasm here. 😉 Technically, we've frozen 2.1, and our attention should be on making sure any fixes to existing features are squared away before considering other things.

For this feature, as you can see from my thoughts above, I don't think it's so cut and dry and ready to go that we can just drop this in before the release. I want to make sure we give ourselves the time to arrive at the most helpful design. In addition, @jodosha's attention has been on our assets fixes and I don't want to rush him with this one either. ☺️

parndt:
`contract do` is what I had before (for many months now, via https://github.com/katafrakt/palaver/blob/d1a31fec7dba6638ec798c419cecf35ce826f28e/lib/palaver/action.rb#L18-L25) and in practice what this meant is that I had a further abstraction over the top of the regular `params do` and had to redeclare `optional(:_csrf_token).maybe(:str?)` otherwise any param that wasn't declared inside `contract do` got wiped. Maybe this was an implementation detail of the way that it was being done, or maybe not, but I wanted to note it. 😄 

Personally, I quite like `params do` followed by any number of `rule` blocks, but open to perspectives too. I guess I don't see the point in adding more abstractions to learn when these concepts exist already and can be hooked into - via DRY.

timriley:
Yeah, I should make this clear: I would like to keep `params do` working no matter what else we do. 

The reason I'm not keen on having many individual `rule` blocks declared directly in the class body of actions is that it suggests that the "rule" behavior is a feature built into the action itself, when the reality is that it's actually just a light-touch integration with extra features offered via dry-validation contracts. 

IMO, the fact that this is using dry-validation is not just a minor implementation detail: we should be making this clear to users, for many reasons:

- so they know why that gem will need to be present
- they'll likely want to no look up dry-validation docs to learn about all the different features they can exercise (I don't think it makes sense for hanami to entirely duplicate those)

In this way, I like the `rules` wrapper block because it is a stronger hint that the rules in aggregate are a part of a _combined thing_, which we can make clear is externally-powered by dry-validation itself. This also provides a contrast to eg. the class-level `before` callbacks and the like, which are, in fact, intrinsic behaviors of Action.

danhealy:
I think I prefer the current implementation for the following reasons:
- Brevity, this implementation avoids the doubly nested `rules do .. rule do` (and similarly, without having to nest `params do`)
- DSL matches that of `Dry::Validation::Contract` to the extent that a future simple integration via `include Dry::Validation::Contract` is compatible.

My complaints with the current implementation:
- Too tightly coupled with the implementation details of `Dry::Validation::Contract` while not providing much additional benefit.  That complaint would go away if it is a stated goal and the expectation to the user is clear that `Hanami::Action` is directly mixing `Dry::Validation::Contract`, over the current more abstract behavior of `hanami-validations`.
- It feels awkward to use `params` as a receiver for an instance of `Dry::Validation::Contract`, rather than something more obvious like `contract`.

Maybe the current DSL + an alternative, explicit receiver for a contract instance would be the best balance?

I don't have a good conceptual judgement about the balance of the ease of use versus potentially surprising behavior of `Hanami::Action` behaving as `Dry::Validation::Contract` only sometimes because of the gem inclusion, but I think that allowing some but not all of `Dry::Validation::Contract`'s DSL is probably not right.

Also, my opinions here include a caveat that they are based on abstract feelings rather than actual usage.

spec/unit/hanami/action/rules_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require "rack"
+
+RSpec.describe Hanami::Action do
+  # Deprecated behavior
```
katafrakt, line 6: I'm not sure why this is "deprecated behavior". This still works, right? Not every action needs rules. Or am I missing something here?

danhealy: Correct!  This comment was from a WIP implementation of `contract do` :)

danhealy, line 6:
```suggestion
```

jodosha:
Hey @danhealy, thanks for this enhancement. 👏 

**TL;DR: Here's my proposal:**

* Keep `params` working as it is.
* Introduce `contract` with `params` and `rule` support.

---

### SemVer compatibility

We need to keep `params` to work as it they are because of Semantic Versioning.

### Why `contract`?

**Reason 1:** Because we can forward the block as it is to `dry-validation`.

For Hanami 3, we can consider **deprecating** `params` in favor of `contract` only.
This move would **reduce the public API** surface of `hanami-controller`, **delegating** to `dry-validation` the entire contract handling.

**Reason 2:** Because `rule` on its own is **meaningless**.
While `params` has a **distinct connotation**, the developers may wonder, "What the heck is a rule?".
It can be a rule for accepting or rejecting MIME Types or an authorization rule... You name it.

But if we wrap it into `contract`, then **the context is explicit**.

---

cc: @timriley @parndt

timriley:
Thanks for checking this out and sharing your preference, @jodosha :)

As @danhealy knows from our (in person! so cool) discussions on this, I didn't have one strong opinion about the exact shape of the API for this, so having you share your thoughts here helps a lot.

I'm happy to go with your suggestion. It's straightforward and provides a single integration point: "contract" builds or accepts a `Dry::Validation::Contract` subclass. That's easy to explain to users, and it will keep the integration with the external dry-validation gem as simple as possible.

The only downside is that for actions that don't need a contract rules, we get this double nesting, which isn't the ideal experience:

```
contract do
  params do
    required(:start_date).value(:date)
    required(:end_date).value(:date)
  end
end
```

An option here could be to keep `params do` within `Action` so that cases like this can just use that directly, rather than `contract`, even though leaving two similar methods in place (for 3.0 or beyond) does muddy the water a bit. In the meantime, however, we have to keep both, so this could be a good opportunity for us to learn from experiences in the wild.

(Another option here could be to rejig dry-validation so that it doesn't need a nested params block like that, but that's an idea for another day!)

@danhealy — since I think we have a clear path now, would you be up for adjusting this PR? Thanks again! ❤️

parndt:
Cool - please let me know when you'd like me to test it out with my app 😄

parndt:
Is this still the direction we're headed?

timriley:
@parndt Yes indeed, I'd be happy to review a revised solution here.

danhealy:
Sorry for letting this sit so long!  The proposal by @jodosha is reasonable, supporting the syntax will be a bit more substantial of a change because it would first depend on modification to `hanami-validations`.

PR #447, opened 2024-06-27
Fix API docs

cllns:
I think we can still go through this again to make sure all the references are solid, but this gets us much closer. 

The big culprit here was the `@api private` in `lib/hanami/action/csrf_protection.rb`. But then I noticed some other things hadn't been updated so I fixed those.


timriley (approved):
Thanks @cllns!

PR #452, opened 2024-07-25
Add :form and :multipart formats

cllns:
These are commonly used for web apps so we should have first-class support for them.

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -15,6 +15,8 @@ module Hanami
       TYPES = {
         txt: "text/plain",
         html: "text/html",
+        form: "application/x-www-form-urlencoded",
+        multipart: "multipart/form-data",
         json: "application/json",
         manifest: "text/cache-manifest",
         atom: "application/atom+xml",


afomera:
I tried this branch out with my application, and was able to resolve an issue I had where my Hanami apps had to define the `form` format for basic form submissions. 🙌

timriley:
Update: I'm taking a slightly different approach. I hope to have some code to share in the next day or two.

The approach is based on storing mime types and content types separately for formats, which allows us to do this:

```ruby
          html: Format.new(
            name: :html,
            mime_type: "text/html",
            content_types: ["application/x-www-form-urlencoded", "multipart/form-data"]
          )
```

timriley:
OK, these entries are now removed as of https://github.com/hanami/hanami-controller/pull/485.

To ensure these content types are accepted for HTML formats, we have the following default format registered:

```ruby
html: Format.new(
  name: :html,
  media_type: "text/html",
  content_types: ["application/x-www-form-urlencoded", "multipart/form-data"]
)
```

What this means is this: if `:html` is configured as an accepted format, then these content types will be permitted, and the response's content type is still set to `text/html` by default.

PR #454, opened 2024-09-02
Allow direct reuse of dry-validation contracts in actions

timriley:


timriley:
@alassek Going to ping you for a review here, since I headed in this direction after your recent prompting. How does feel to you? Does it give you what you'd hoped for?

timriley:
Merging this one now. We can always make further adjustments later.

PR #460, opened 2024-12-31
rack 3 support

kyleplump:
updates to Rack 3. one part of many gem updates

[list of changes](https://github.com/hanami/router/pull/277)

### DIFF
--- a/hanami-controller.gemspec
+++ b/hanami-controller.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.metadata["rubygems_mfa_required"] = "true"
   spec.required_ruby_version = ">= 3.1"
 
-  spec.add_dependency "rack", "~> 3.0"
+  spec.add_dependency "rack", "~> 3.1"
   spec.add_dependency "hanami-utils", "~> 2.2"
   spec.add_dependency "dry-configurable", "~> 1.0", "< 2"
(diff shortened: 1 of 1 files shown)


hanami-controller.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.metadata["rubygems_mfa_required"] = "true"
   spec.required_ruby_version = ">= 3.1"
 
-  spec.add_dependency "rack", "~> 2.0"
+  spec.add_dependency "rack", "~> 3.1"
```
timriley, line 23: I think we'll want to make this `">= 2.0"` like https://github.com/hanami/router/pull/277?

We should add a matrix to CI here to test on both Rack 2 & 3.

timriley (commented):
I think we need a small change here, to the rack version specification (we still want to allow Rack 2).

bkuhlmann:
I'd like this too. Can this be folded in if there is no strong reason for not running on Rack 3.0.0?

timriley:
@bkuhlmann Yes, I'm keen to get this in. All the related Rack 3 PRs will be merged together as soon as they're ready. @kyleplump has a few changes to sort out [in the main PR](https://github.com/hanami/router/pull/277#pullrequestreview-2726746467) before that can be done.

Once these are ready, we'll be preparing a Hanami 2.3 release to get them out the door.

PR #471, opened 2025-06-03
Make use to #media_type instead of #content_type to avoid false negatives (e.g., multipart/form-data; boundary=...)

wuarmin:
Hey :wave: 

this PR would fix https://discourse.hanamirb.org/t/how-do-you-set-up-a-file-upload-route-correctly-for-a-json-rest-api/1222/3

thanks

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -222,11 +222,11 @@ module Hanami
         # @since 2.0.0
         # @api private
         def enforce_content_type(request, config)
-          content_type = request.content_type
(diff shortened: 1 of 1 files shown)


timriley:
This looks great, @wuarmin! Thanks for doing it!

Could you please add one test to demonstrate this working? Your exact example from your forum post would be great.

wuarmin:
@timriley,
perhaps you no longer have this important bug fix on your radar. I just wanted to remind you, as file upload is currently being discussed. It might save you some time.

PR #481, opened 2025-09-22
Add subdomains and subdomain methods to Hanami::Action::Request

wout:
Here's the pull request as I proposed in the Discord. I'm not familiar with the annotations in the comments, but I tried to follow the existing patterns.

One thing I was still wondering about, should the TLD length be configurable at a higher level so that it's not necessary to pass the value explicitly for domains with multiple TLDs? If so, where, and how?

### DIFF
--- a/spec/unit/hanami/action/config_spec.rb
+++ b/spec/unit/hanami/action/config_spec.rb
@@ -70,11 +70,6 @@ RSpec.describe Hanami::Action::Config do
       config.default_tld_length = 2
       expect(config.default_tld_length).to be(2)
     end
-
-    it "rejects nil values" do
-      config.default_tld_length = nil
-      expect(config.default_tld_length).to be(1)
-    end
   end
 
   describe "#cookies" do


lib/hanami/action.rb
```diff
@@ -62,6 +62,7 @@ def self.gem_loader
     setting :formats, default: Config::Formats.new, mutable: true
     setting :default_charset
     setting :default_headers, default: {}, constructor: -> (headers) { headers.compact }
+    setting :default_tld_length, default: 1, constructor: -> (length) { length || 1 }
```
timriley, line 65:
```suggestion
    setting :default_tld_length, default: 1
```

I don't think we need to make this so defensive, at least not to start with.

lib/hanami/action.rb
```diff
@@ -440,7 +441,7 @@ def session_enabled?
     # @since 2.0.0
     # @api private
     def build_request(**options)
-      Request.new(**options)
+      Request.new(default_tld_length: config.default_tld_length, **options)
```
timriley, line 444: Whoops, I gave you bad guidance on this one. This arg should be passed down in `#call`, as one of these:

```ruby
        request  = build_request(
          env: env,
          params: params,
          session_enabled: session_enabled?
        )
```

lib/hanami/action/config.rb
```diff
@@ -117,6 +117,24 @@ def format(*formats)
...
+      #   subdomain(s) in `Request#subdomains`.
+      #
+      #   Defaults to 1.
+      #
+      #   @example
+      #     # For *.example.com
+      #     config.default_tld_length = 1
+      #
+      #     # Or for *.example.co.uk
+      #     config.default_tld_length = 2
```
timriley, line 132: These are very helpful examples, thank you!

lib/hanami/action/request.rb
```diff
@@ -26,11 +26,12 @@ class Request < ::Rack::Request
 
       # @since 2.0.0
       # @api private
-      def initialize(env:, params:, session_enabled: false)
+      def initialize(env:, params:, default_tld_length:, session_enabled: false)
```
timriley, line 29:
```suggestion
      def initialize(env:, params:, default_tld_length: 1, session_enabled: false)
```

Let's add a default arg here, just so this class makes sense on its own.

I think this will mean you can avoid having to manually pass `default_tld_length: 1` when constructing the requests in a bunch of the tests you changed.

timriley (commented):
This is all looking great, great stuff @wout! I've left just a couple small recommendations. After these are done we can get this merged :) Thank you again!

wout:
I think this is it @timriley, but please let me know if there's anything else that you'd like to see changed. :)

Issue #482, opened 2025-09-28
MIME/Format handling should check extensions if an `Accept` header is absent


timriley:
This would be helpful for developers setting up `.json` routes and wanting to casually test them in local dev.

PR #483, opened 2025-09-28
MIME/Format handling should check extensions and use path format.

afomera:
👋 First time contributor to Hanami here so tell me how we'd be best to improve this code to make it better for all and maintainable if it's not.

In terms of disclosure: I did use AI to help determine where to put the code as I was unfamiliar with the codebase but I did review every line suggested and tested it in my Hanami application manually. If this isn't allowed please do let me know and I'll revisit it manually 🙏.

Resolves: https://github.com/hanami/controller/issues/482

This code allows us to define in our actions:

```ruby

module Bookshelf
  module Actions
    module Books
      class Index < Bookshelf::Action
        format :json, :html, :xml

        def handle(request, response)
          if response.format == :xml
            response.body = "<books></books>"
            return
          end

          if response.format == :json
            response.body = { OK: true }.to_json
            return
          end

          response.render(
            view,
            page: request.params[:page] || 1,
            per_page: request.params[:per_page] || 5
          )
        end
      end
    end
  end
end
```

This allows smooth handling of response formats from the same action, which allows me to make a simple action to list books (or whatever I'm doing), in html, json and xml if I'm feeling spicy.

With a route defined for .json / .xml and /books to the same action this allows me to just write a different response handler based on the automatically set response.format from the path.

### DIFF
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -254,7 +254,7 @@ module Hanami
           return nil if extension.nil? || extension.empty?
 
           # Remove the leading dot and return
-          extension[1..-1]
+          extension[1..]
         end
 
         # Returns the format based on path extension
--- a/lib/hanami/action/mime.rb
+++ b/lib/hanami/action/mime.rb
@@ -233,6 +233,45 @@ module Hanami
 
         private
 
+        # Extracts the file extension from a path
+        #
+        # @param path [String] the request path
+        # @return [String, nil] the file extension without the dot, or nil if no extension
+        #
+        # @example
+        #   extract_extension("/users/123.json") # => "json"
+        #   extract_extension("/users/123") # => nil
+        #
+        # @api private
+        def extract_extension(path)
+          return nil if path.nil? || path.empty?
+
+          # Remove query string and fragment if present
+          clean_path = path.split("?").first.split("#").first
+
+          # Extract extension
+          extension = File.extname(clean_path)
+          return nil if extension.nil? || extension.empty?
+
+          # Remove the leading dot and return
+          extension[1..-1]
+        end
+
+        # Returns the format based on path extension
+        #
+        # @param path [String] the request path
+        # @return [Symbol, nil] the format symbol, or nil if no matching format
+        #
+        # @api private
+        def format_from_path_extension(path)
+          extension = extract_extension(path)
+          return nil unless extension
+
+          # Convert extension to symbol and check if it's a known format
+          format = extension.to_sym
+          TYPES.key?(format) ? format : nil
+        end
+
         # @since 2.0.0
         # @api private
         def accepted_mime_type?(mime_type, config)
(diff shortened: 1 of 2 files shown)


afomera:
Thank you @alassek! That's a super helpful and detailed review 🙌 

I'm a first-time user of Hanami and coming from another framework, it certainly took me what I'd consider far too long to figure out how to get my Action to respond with json and html 😢, if I wasn't highly motivated to stick with it and learn -- I would have hit the escape hatch and bailed and had the idea that "Hanami can't do something that's trivial in other frameworks" (which, would be wrong, but I spent about 2 hours last night trying to figure out this functionality, which is like an hour and half longer than my attention span would have permitted otherwise)

Now I'm obviously not familiar with the vision (is there a doc somewhere outlining it?) for Hanami in the short and long term so take that with a grain of salt. from what I understand in ~12 hours of using Hanami seriously (and being very frustrated while learning), there's an emphasis on explicit configuration.

I am concerned a middleware will be buried and essentially be unused until you magically find it in the docs / code. If it _should_ be a middleware then I would ideally think it should be enabled by default with the ability to opt out. 

I went through a few attempts on my own for how I thought it could be solvable:

1. tried passing `request.params[:format]` to the response.render call in the action/view class to get it to render a specific template (lol rails habits die hard). When this didn't work:
2. Tried to look in request for the `.format` that might have been there to look if I could tell if the request came in by books.json or whatnot, but it's not available so this didn't let me do some if statements in my action.
3. Checked if the router allowed me to specific routes with extensions to go to specific actions or if I could define `format: :json` Ala rails style to specifically only respond to those formatted requests with specific json focused actions.

obviously none of those worked or were correct though. I would have loved to have been able to do something similar to the respond_to block in Rails controllers after defining the .json route to the same action and been on my way.

Because you have to setup a route with the extension in the router it felt like it was less-likely to conflict, and likely would auto-infer for 90% of cases where someone is looking at the path_info in their action and setting the response.format = :json themselves if their route was '.json'. 

Sinatra provides 'conditions' for their routes: https://sinatrarb.com/intro.html (search Conditions) and see you can define a `provides` which looks at the Accept header for that stuff. Above they also allow 'format' for their router. Rails just kinda defaults to sending / setting the `.format` to the accept header or the extension and then other non-ruby frameworks I played with in the past feel similarly IIRC (it's been awhile). I guess I've been spoiled and used to frameworks that bend the HTTP standard 😅. 

Sorry for the book, I'm happy to adjust to a middleware if that's what we think will be best. Just wanted to outline my thinking some from a first-time user incase it's helpful.

alassek:
> Checked if the router allowed me to specific routes with extensions to go to specific actions or if I could define format: :json Ala rails style to specifically only respond to those formatted requests with specific json focused actions.

We should probably document this, but the router does support expressing this:

https://github.com/hanami/router/blob/main/spec/integration/hanami/router/recognition_spec.rb#L322-L335

```ruby
get '/test(.:format)' # optional format
get '/test.:format`  # required format
```

afomera:
I've been thinking on the best approach to handle this and I agree my first approach was probably not ideal for the framework core. 

I'm going to close this for now, if someone one else picks this issue up before me that's ok!

PR #487, opened 2025-10-18
Add support for handling exceptions defined in config as strings

krzykamil:
As per https://guides.hanamirb.org/v2.2/actions/exception-handling/ we are adding support for defining the exception classes as strings, the reason for this provided here: https://github.com/hanami/hanami-controller/issues/432
Example usage
```ruby
  handle_exception "RecordNotFound" => 404
  handle_exception "StandardError" => :handle_error
  
  def handle(_req, _res)
    raise RecordNotFound.new
  end

  private 

  def handle_error(request, response, exception)
    smart_stuff
  end
end
```

In case of an non-existing exception being provided in the string, we rescue from NameError that would have been thrown in those cases

- [ ] Will need to add this to 2.3 documentation if accepted

### DIFF
--- a/spec/support/fixtures.rb
+++ b/spec/support/fixtures.rb
@@ -615,6 +615,23 @@ class StringExceptionWithHandlerAction < Hanami::Action
   end
 end
 
+class NonExistentExceptionNameAction < Hanami::Action
+  config.handle_exception "NonExistentException" => 404
+
+  def handle(_req, _res)
+    raise RecordNotFound.new
+  end
+end
+
+class MixedExistingAndNonExistentAction < Hanami::Action
+  config.handle_exception "NonExistentException" => 400
+  config.handle_exception RecordNotFound => 404
+
+  def handle(_req, _res)
+    raise RecordNotFound.new
+  end
+end
+
 class ParamsAction < Hanami::Action
   def handle(req, res)
     params = req.params.to_h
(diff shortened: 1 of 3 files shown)


timriley:
@krzykamil Thanks for dong this! Looking at the code, this will take the string version of the exception and immediately try and constantize it, is that right?

So the goal here was to make it so that you don't have to do e.g. `require "rom"` at the top of an action file simply to be able to add `ROM::TupleCountMismatchError` to the action's list of handled exceptions. It feels a bit "wrong" to have to load a database library inside your app's HTTP layer in order to do this.

With this PR as it is, I _think_ that you'd still need to put `require "rom"` at the top of the file for this to work?

To avoid this, do you think you could look at adjusting this so it avoids constantizing the exception class strings entirely, or at least defers it until the latest possible moment?

So maybe we could change it so that `config.handled_exceptions` actually holds these strings (not constantized), and then change `Action#exception_handler` so that it does something like this?

```ruby
return handler if exception.is_a?(Class) ? exception.is_a?(exception_class) : exception.to_s == class.name
```

Or something like that? This way, we can avoid having to do those random `require`s at the top of the file just to rescue from an error.

timriley:
👋🏼 hi again @krzykamil, looks like #488 was doing the thing I suggested just above, so I've merged it in, which means we can consider this feature all done. Thank you again for putting in this effort ❤️

PR #488, opened 2025-10-19
Allow `handle_exception` to accept string class name

### LINKED ISSUE #432: handle_exception should allow exception names as strings
It would be nice to be able to rescue from anticipated errors without needing to _require the gem_ at the top of the action in order to make that error constant available at the time the action is defined.

This would also make it more consistent with our new `render_error_responses` setting:

```ruby
    setting :render_error_responses, default: Hash.new(:internal_server_error).merge!(
      "Hanami::Router::NotAllowedError" => :not_found,
      "Hanami::Router::NotFoundError" => :not_found,
    )
```

mathewdbutton:
This PR attempts to implement changes requested in https://github.com/hanami/hanami-controller/issues/432

Previously an error class would have to be supplied, which might mean requiring a gem in that action.

This change allows for a string to be defined instead, removing the need for requiring gems for error classes.

Fixes #432

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -433,7 +433,12 @@ module Hanami
     # @api private
     def exception_handler(exception)
       config.handled_exceptions.each do |exception_class, handler|
-        return handler if exception.is_a?(exception_class)
+        case exception_class
+        when String
+          return handler if exception.class.name == exception_class # rubocop:disable Style/ClassEqualityComparison
+        else
+          return handler if exception.is_a?(exception_class)
+        end
       end
 
       nil
--- a/spec/support/fixtures.rb
+++ b/spec/support/fixtures.rb
@@ -116,6 +116,14 @@ class ErrorCallAction < Hanami::Action
   end
 end
 
+class ErrorCallWithStringClassName < Hanami::Action
+  config.handle_exception "RuntimeError" => 500
+
+  def handle(_req, _res)
+    raise
+  end
+end
+
 class MyCustomError < StandardError; end
 
 class ErrorCallFromInheritedErrorClass < Hanami::Action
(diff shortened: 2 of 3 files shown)


lib/hanami/action.rb
```diff
@@ -433,7 +433,12 @@ def enforce_accepted_media_types(request)
     # @api private
     def exception_handler(exception)
       config.handled_exceptions.each do |exception_class, handler|
-        return handler if exception.is_a?(exception_class)
+        case exception_class
+        when String
+          return handler if exception.class.name == exception_class # rubocop:disable Style/ClassEqualityComparison
```
mathewdbutton: This was to avoid
```
lib/hanami/action.rb:438:39: C: Style/ClassEqualityComparison: Use instance_of? instead of comparing classes. (https://rubystyle.guide#instance-of-vs-class-comparison)
          return handler if exception.class.name == exception_class
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Which I don't _think_ we want to do, we want to make sure that the two strings match

mathewdbutton:
HAH! I should have checked this hadn't been already snaffled up https://github.com/hanami/hanami-controller/pull/487. OOPS!

lib/hanami/action.rb
timriley: Yeah, I think we actually want to support either a string _or_ a real exception class being provided. Would you mind adjusting it to do that, @mathewdbutton?

timriley:
@mathewdbutton Would you mind updating the guide for 2.3 to include this? Relevant file is https://github.com/hanami/guides/blob/main/content/v2.3/actions/exception-handling.md. Thanks :)

mathewdbutton:
On it!

PR #490, opened 2025-10-25
Gracefully handle the rename of Status 422 Unprocessable Content

alassek:
Rack 3 changed the symbolic name of 422 from `:unprocessable_entity` to `:unprocessable_content`.

This follows from the rename in the HTTP spec. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422

### DIFF
--- a/lib/hanami/http/status.rb
+++ b/lib/hanami/http/status.rb
@@ -21,7 +21,10 @@ module Hanami
       #
       # @since 2.0.2
       # @api private
-      SYMBOLS = ::Rack::Utils::SYMBOL_TO_STATUS_CODE
+      SYMBOLS = ::Rack::Utils::SYMBOL_TO_STATUS_CODE.merge(
+        unprocessable_content: 422,
+        unprocessable_entity: 422
+      )
 
       # Return a status for the given code
       #
--- a/spec/support/version_matcher.rb
+++ b/spec/support/version_matcher.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class VersionMatcher
+  attr_reader :spec
+
+  def initialize(spec)
+    @spec = Gem.loaded_specs[spec]
+  end
+
+  def matches?(version)
+    requirement(version).satisfied_by? spec.version
+  end
+
+  def excludes?(version)
+    !matches?(version)
+  end
+
+  def to_proc
+    method(:excludes?).to_proc
+  end
+
+  private
+
+  def requirement(version)
+    return version unless version.is_a?(String)
+
+    Gem::Requirement.create version.split(", ")
+  end
+end
+
+RSpec.configure do |config|
+  config.filter_run_excluding rack: VersionMatcher.new("rack").to_proc
+end
--- a/spec/unit/hanami/http/status_spec.rb
+++ b/spec/unit/hanami/http/status_spec.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+RSpec.describe Hanami::Http::Status do
+  subject(:status) { described_class }
+
+  describe ".for_code" do
+    it "accepts an Integer status code" do
+      expect(status.for_code(200)).to eq([200, "OK"])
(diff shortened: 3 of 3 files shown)


lib/hanami/http/status.rb
```diff
@@ -21,7 +21,13 @@ class Status
       #
       # @since 2.0.2
       # @api private
-      SYMBOLS = ::Rack::Utils::SYMBOL_TO_STATUS_CODE
+      SYMBOLS = ::Rack::Utils::SYMBOL_TO_STATUS_CODE.dup
+
+      if Gem.loaded_specs["rack"].version.version.start_with?("3")
```
alassek: Simplest way I could think to determine this with a minimum of allocations

afomera: I was thinking we could just use `Rack::VERSION` to check against rather than dropping this far, but rack 2's version const is... weird.

I think I might change that second `version` call for `to_s` as `Gem::Version` does respond with a string version

```rb
Gem.loaded_specs["rack"].version.to_s.start_with?("3")
```


```
± be hanami console
forked[development]> Gem.loaded_specs["rack"].version.to_s.start_with?("3")
=> true
forked[development]> Gem.loaded_specs["rack"].version.to_s
=> "3.2.3"
```

and in rack 2

```
± be hanami console
forked[development]> Rack::VERSION
=> [1, 3]
forked[development]> Rack
=> Rack
forked[development]> Rack::VERSION.to_s
=> "[1, 3]"
forked[development]> Gem.loaded_specs["rack"].version.version
=> "2.2.20"
forked[development]> Gem.loaded_specs["rack"].version
=> Gem::Version.new("2.2.20")
forked[development]> Gem.loaded_specs["rack"].version.to_s
=> "2.2.20"
forked[development]> Gem.loaded_specs["rack"].version.to_s
=> "2.2.20"
forked[development]> Gem.loaded_specs["rack"].version.class
=> Gem::Version
forked[development]> Gem.loaded_specs["rack"].version.version
=> "2.2.20"
forked[development]> Gem.loaded_specs["rack"].version.to_s
=> "2.2.20"
```

lib/hanami/http/status.rb
```diff
@@ -21,7 +21,13 @@ class Status
...
+        SYMBOLS[:unprocessable_entity] = 422
+      else
+        SYMBOLS[:unprocessable_content] = 422
```
aaronmallen: How do we feel about keeping `:unprocessable_entity` in **addition** to the new symbol for backwards compatibility?

lib/hanami/http/status.rb
timriley: We have a simple check for this that you can reuse: `Hanami::Action.rack_3?`

(See `lib/hanami/action/rack_utils.rb` and feel free to grep for it across the repo to see how we use it).

lib/hanami/http/status.rb
timriley: I think that because we've [documented `unprocessable_entity`](https://guides.hanamirb.org/v2.1/actions/status-codes/), it would be good for us to maintain it in the future.

I think for our 2.3 docs, we should show _both_ `unprocessable_content` and `unprocessable_content` as working for 422, and should update this PR to work with both, for either version of Rack.

This feels important during this Rack v2/v3 transitional phase, which I expect we might keep for ~a year or so.

I _think_ this might also simplify this PR? We'd just merge unprocessable_content and unprocessable_content keys into our `SYMBOLS` hash regardless of the version of rack we see being used?

timriley (commented):
Thank you so much for doing this, @alassek! This will take away one little hitch for our users and hopefully make their upgrade nicer :)

I've left a thought via a comment about a change in approach. Keen to hear what you think.

spec/support/version_matcher.rb
timriley: This is clever! 😄

timriley (approved):
Thanks @alassek, looks great!

PR #494, opened 2025-11-27
Update yardoc to correct type

keegankb93:
Original type is Fixnum, but should be Integer or Symbol which is used to determine the http status.

https://github.com/hanami/hanami-controller/issues/493#issue-3665970596


timriley:
That would be wonderful, thank you @keegankb93. Happy to help with your future contributions :) You can always find me in our chat or forum or even by email if that's easiest. Enjoy your vacation!

PR #495, opened 2025-12-05
Fix `handle_exception` string class name error

sidane:
Related to #488 which allowed for passing the string name of a class to handle_exception.

However, when passing more than one string class name like this:

```ruby
handle_exception(
  "MyException" => 500,
  "MyOtherException" => 501
)
```

`Hanami::Action::Config#handle_exception` errors with:

```ruby
Failure/Error: ex1.ancestors.include?(ex2) ? -1 : 1

NoMethodError:
  undefined method 'ancestors' for an instance of String
```

This fixes that error by effectively skipping the sorting of exception classes when any of the two classes being compared is a string.


timriley:
Shipped in v2.3.1 :)

PR #508, opened 2026-04-24
Fix url params getting removed

cllns:
Fixes a bug I found introduced in #500: when a JSON body was being parsed, params from the url were being stripped.

The spec I wrote fails for 'main', but passes for v2.3.1 and after this change, confirming this was a regression.

I also added an analogous regression test for multipart form submissions


timriley (approved):
Great catch, thank you @cllns! I'm going to merge this right away, since I kind of want to poke at some refactors to make things a bit clearer.

PR #509, opened 2026-04-24
Reinstate inheritance integration spec

cllns:
`resources` used to be available in router, now it's in hanami/hanami

So this is a suitable replacement!

I also moved the module with its classes into the spec. I think of Fixtures as a place for shared things. It's easier to read this spec in isolation without having to go to fixtures file.


spec/integration/inheritance_spec.rb
```diff
@@ -2,15 +2,78 @@
...
+    before :authenticate!
+
+    private
+
+    def authenticate!(*, res)
+      res[:authenticated] = true
+    end
+  end
+
+  module Controllers
```
timriley: Future TODO: remove all instances of the "Controller" terminology even from our test class namespaces (I felt that was too much to bite off with the initial rename PR).

timriley (approved):
Good to have this back, thank you!

This never really needed to be tied to resource-based organisation of actions in the first place, anyway. This is just straightforward testing of inheritance and its impact on callbacks/super, etc.

timriley:
@cllns Regarding fixtures — one of the things we avoid in this test suite by using fixtures is class names leaking between tests. Right now we don't have any machinery to clean up constants after tests, etc., so if we happened to want to define `InheritanceSpec::Action` in _another_ example group, that would end up causing problems, I think.

Maybe we could use anonymous classes here to avoid that?

cllns:
Ah great point Tim. I switched to use anonymous classes. This require re-working the spec since we were passing the class name along, which anonymous classes don't have. Instead I used #object_id which is a suitable replacement.

PR #510, opened 2026-04-24
Reinstate fullstack spec & refactor it

cllns:
It may be easier to review this PR commit-by-commit.

1. 40e8ff5 - Removed pending, updated Renderer to not render HEAD requests
2. 9b657e1 - Moved the FullStack fixture into the spec. It's not re-used elsewhere so this makes the spec self-contained
3. 6e7e1f3 - Added a spec for a missing coverage: POST to /poll/2
4. 0bc1316 - Changed Renderer to render JSON instead of just a string. The expectations on the string were brittle. 
    - I also changed the expectations to match the entire parsed response hash instead of just a few select items. This means the expectations reveal less intent, but it means it's harder to miss something. I think it's always obvious what we're trying to test, and the other values are just empty, so they're easy to scan over. I'm open to changing this if we want to go to expectations that are more self-evidently meaningful, but I'd rather err on over-checking then opening the possibility for under-checking.
5. e155f5d - Rename Renderer to ResponseSerializer since that's more accurate now. Move it into full_stack_spec. It was used a couple other places (session tests) but it wasn't needed there. Those specs have some pending tests but they're unrelated (they just check headers). Besides, they'd benefit from more narrowly targeted files since ResponseSerializer has a lot going on.


cllns:
Fixed merge conflicts and got CI green, ready for review

timriley (approved):
This is great, @cllns! Ever since I started working on this codebase I've wanted to deal with these sprawling fixtures. Reenabling this test and inlining its classes is a great step :)

PR #514, opened 2026-06-06
Performance: Reduce hot path allocations

cllns:
A batch of focused, low-risk per-request allocation cuts on `Hanami::Action#call`. Each commit is a single, attributable optimization (with the allocation delta in its commit message); the cumulative effect is roughly **half the allocations** and **~1.5–2.5× the throughput** of every action invocation.

I used Claude Code heavily to find the optimizations, assess their risk, and benchmark. I also had it draft the text of this PR, which I've modified _heavily_. I wouldn't have had the time to do this PR so extensively and well-documented without AI assistance. 

## Cumulative results

**Speed**

| Scenario | IPS |  | μs/call |  | Speedup |
|---|---|---|---|---|---|
| `Action#call` | before | after | before | after |  |
| • no formats | 123.5k | 302.6k | 8.09 | 3.31 | **2.45×** |
| • formats + `Accept` | 73.2k | 113.5k | 13.65 | 8.81 | **1.55×** |

**Memory**

| Scenario | Allocations |  | Allocations saved | Bytes |  | Memory saved |
|---|---|---|---|---|---|---|
| `Action#call` | before | after |  | before | after |  |
| • no formats | 51 | 16 | **69%** | 4.54 kB | 1.70 kB | **63%** |
| • formats + `Accept` | 95 | 61 | **36%** | 6.70 kB | 3.94 kB | **41%** |

Ruby 4.0.4, M1 Pro. 

## Per-commit attribution

Each commit passes test suite. Allocation savings per call, measured via `memory_profiler`:

| SHA | Commit | Allocations reduced | Notes |
|---|---|---|---|
| 3c8f171 | Avoid splat allocations on the action hot path | −4 | Replace `**options`/`*args` in `build_request`, `build_response`, and the callback runners with explicit kwargs. We could also just do `...` instead of listing all the kwargs, which is the same allocation reduction, but I felt this is more readable to know exactly what's passed through. |
| 4fb3059 | Skip DefaultContract machinery when no contract is configured | −8 | When neither `.params` nor `.contract` is used, deep-symbolize the raw params directly instead of routing through `DefaultContract.call` (and reuse a frozen empty hash when there's nothing to parse). The `DefaultContract` class becomes unused and is removed. |
| 5e3a311 | Skip Rack::Request allocation when no params to extract | −4 | Inspect the env directly to decide which param sources are present and only build a `Rack::Request` when there's actually a query string or form body to parse. Requests with no params short-circuit to a frozen empty hash. This adds considerable complexity (and therefore requires disabling some cops) but I think it's better than adding indirection via breaking out helper methods for this. Open to refactoring though, if desired. |
| 27dcab2 | Precompute response charset, skip Rack::MediaType parse per request | −10 | The single largest per-call win. Compute the charset once per action from config and pass it to each `Response`, so it no longer re-parses the content type via `Rack::MediaType.params` on every request. Response keeps the parse as a fallback for callers that don't pass an explicit charset. It's almost always UTF8 anyway. |
| f8fefcc | Cache default response content type and format per action | −7 (−8 total with no formats) | Precompute the content type + format used when a request has no usable `Accept` header. `Action#finish` reuses the cached format whenever `res.content_type` still matches the default, avoiding `Mime.extract_media_type` re-parsing on every request. |
| 3058417 | Skip redundant headers.dup in Response#initialize | −1 | `Rack::Response#initialize` already copies the headers into its own internal store on both Rack 2.2.16+ and 3.x, so the defensive `headers.dup` we were doing was a redundant intermediate copy. A regression spec asserts `default_headers` stays unpolluted across requests so any future Rack regression would fail loudly instead of leaking response state. |
| 3eefda0 | Use frozen EMPTY_BODY when clearing Response#body | −1 (HEAD/204/304 only) | When `Response#body=` is called with `nil` or the empty-body sentinel, used by `Action#finish` for HEAD requests and bodyless statuses (204, 304), assign the frozen `EMPTY_BODY` constant directly instead of allocating a duped empty array. No effect on the standard "write a body" path. |

## Behavioral notes

These are tightly scoped allocation cleanups, not API changes. Two are worth calling out explicitly:

- **`Rack::Request` skip changes when `Rack::Request#POST` runs.** The old code called `Rack::Request#POST` for any body without `ACTION_BODY_PARAMS`; the new code only does so when the content-type is `application/x-www-form-urlencoded` or `multipart/form-data`. For non-form bodies (e.g. raw JSON whose format wasn't accepted), Rack's `POST` would have returned `{}` and we'd merge nothing anyway, but a side effect is that we no longer consume `rack.input` for those requests. All 99 params + body-parsing specs pass; flagged for visibility.

- **`headers.dup` removal** undoes a guard explicitly added in #245 ("Remove global state"). The safety now depends on Rack's `Response#initialize` continuing to copy the headers hash, which is effectively guaranteed by Rack's API contract (a `Response` must own its headers to function). Verified empirically on Rack 2.2 and 3.2, a new regression spec catches any future breaking changes to this assumption.

## Reproducing the numbers

Two representative actions exercised via `benchmarks/action_call_ips.rb` and `benchmarks/action_call_memory.rb`, scripts not included in this diff, but shown below.

Each action is instantiated once and reused, matching the container-managed singleton pattern used in production Hanami apps.

<details>
<summary><code>benchmarks/action_call_ips.rb</code></summary>

```ruby
# frozen_string_literal: true

require "hanami/action"
require "benchmark/ips"

class HelloAction < Hanami::Action
  config.formats.accept :json
  config.default_headers = {"X-Frame-Options" => "DENY"}

  def handle(_req, res)
    res.body = "hello"
  end
end

class MinimalAction < Hanami::Action
  def handle(_req, res)
    res.body = "hello"
  end
end

MINIMAL_ACTION = MinimalAction.new
HELLO_ACTION = HelloAction.new

ENV_NO_ACCEPT = {
  "REQUEST_METHOD" => "GET",
  "PATH_INFO" => "/",
  "QUERY_STRING" => ""
}.freeze

ENV_WITH_ACCEPT = ENV_NO_ACCEPT.merge("HTTP_ACCEPT" => "application/json").freeze

Benchmark.ips do |x|
  x.report("Action#call (no formats)") do
    MINIMAL_ACTION.call(ENV_NO_ACCEPT.dup)
  end

  x.report("Action#call (with formats + Accept)") do
    HELLO_ACTION.call(ENV_WITH_ACCEPT.dup)
  end

  x.compare!
end
```

</details>

<details>
<summary><code>benchmarks/action_call_memory.rb</code></summary>

```ruby
# frozen_string_literal: true

require "hanami/action"
require "memory_profiler"

class HelloAction < Hanami::Action
  config.formats.accept :json
  config.default_headers = {"X-Frame-Options" => "DENY"}

  def handle(_req, res)
    res.body = "hello"
  end
end

class MinimalAction < Hanami::Action
  def handle(_req, res)
    res.body = "hello"
  end
end

MINIMAL_ACTION = MinimalAction.new
HELLO_ACTION = HelloAction.new

ENV_NO_ACCEPT = {
  "REQUEST_METHOD" => "GET",
  "PATH_INFO" => "/",
  "QUERY_STRING" => ""
}.freeze

ENV_WITH_ACCEPT = ENV_NO_ACCEPT.merge("HTTP_ACCEPT" => "application/json").freeze

# Warm up class-level state before measuring per-call allocations.
MINIMAL_ACTION.call(ENV_NO_ACCEPT.dup)
HELLO_ACTION.call(ENV_WITH_ACCEPT.dup)

times = Integer(ENV.fetch("TIMES", 1))

puts "== Action#call (no formats) #{times}× =="
report = MemoryProfiler.report do
  times.times { MINIMAL_ACTION.call(ENV_NO_ACCEPT.dup) }
end
report.pretty_print(scale_bytes: true, detailed_report: false)

puts
puts "== Action#call (with formats + Accept) #{times}× =="
report = MemoryProfiler.report do
  times.times { HELLO_ACTION.call(ENV_WITH_ACCEPT.dup) }
end
report.pretty_print(scale_bytes: true, detailed_report: false)
```

</details>

## Future work
- We can add a per-action cache of the negotiated content type, which should reduce the `formats + Accept` case by ~25 more allocations.
- We're duplicating params parsing here and in `hanami-router`: both libraries parse the query string, both parse form bodies, and both deep-symbolize parsed JSON bodies. When `router.params` is set (which `hanami-router` always does), we can trust it as the canonical pre-merged params bag and skip our own `Rack::Request` allocation entirely. This should save ~20–30 additional allocations per request with a non-empty query string or a form body, when running behind `hanami-router`.

### DIFF
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -474,16 +474,16 @@ module Hanami
     #
     # @since 2.0.0
     # @api private
-    def build_request(**options)
-      Request.new(**options)
+    def build_request(env:, params:, session_enabled:, default_tld_length:)
+      Request.new(env:, params:, session_enabled:, default_tld_length:)
     end
 
     # Hook to be overridden by `Hanami::Extensions::Action` for integrated actions
     #
     # @since 2.0.0
     # @api private
-    def build_response(**options)
-      Response.new(**options)
+    def build_response(request:, config:, content_type:, env:, headers:, session_enabled:, view_options: nil)
+      Response.new(request:, config:, content_type:, env:, headers:, session_enabled:, view_options:)
     end
 
     # @since 0.2.0
@@ -535,15 +535,15 @@ module Hanami
 
     # @since 0.1.0
     # @api private
-    def _run_before_callbacks(*args)
-      config.before_callbacks.run(self, *args)
+    def _run_before_callbacks(req, res)
+      config.before_callbacks.run(self, req, res)
       nil
     end
 
     # @since 0.1.0
     # @api private
-    def _run_after_callbacks(*args)
-      config.after_callbacks.run(self, *args)
+    def _run_after_callbacks(req, res)
+      config.after_callbacks.run(self, req, res)
       nil
     end
 
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -293,6 +293,13 @@ module Hanami
     # @api private
     private attr_reader :contract
 
+    # Resolved response charset, computed once from config. Passed to each {Response} so it
+    # doesn't have to re-parse the content type's charset on every request.
+    #
+    # @since x.x.x
+    # @api private
+    private attr_reader :default_charset
+
     # Returns a new action
     #
     # @since 2.0.0
@@ -300,6 +307,7 @@ module Hanami
     def initialize(config: self.class.config, contract: nil)
       @config = config.finalize!.to_data
       @contract = contract || config.contract_class&.new # TODO: tests showing this overridden by a dep
+      @default_charset = @config.default_charset || DEFAULT_CHARSET
       freeze
     end
 
@@ -343,6 +351,7 @@ module Hanami
           request: request,
           config: config,
           content_type: Mime.response_content_type_with_charset(request, config),
+          charset: default_charset,
           env: env,
           headers: config.default_headers,
           session_enabled: session_enabled?
@@ -482,8 +491,9 @@ module Hanami
     #
     # @since 2.0.0
     # @api private
-    def build_response(request:, config:, content_type:, env:, headers:, session_enabled:, view_options: nil)
-      Response.new(request:, config:, content_type:, env:, headers:, session_enabled:, view_options:)
+    def build_response(request:, config:, content_type:, env:, headers:, session_enabled:, charset: nil,
+                       view_options: nil)
+      Response.new(request:, config:, content_type:, charset:, env:, headers:, session_enabled:, view_options:)
     end
 
     # @since 0.2.0
--- a/lib/hanami/action.rb
+++ b/lib/hanami/action.rb
@@ -23,7 +23,7 @@ module Hanami
   #   require "hanami/action"
   #
   #   class Show < Hanami::Action
-  #     def handle(req, res)
+  #     def handle(request, response)
   #       # ...
   #     end
   #   end
@@ -159,7 +159,7 @@ module Hanami
     #     class Show < Hanami::Action
     #       before :authenticate, :set_article
     #
-    #       def handle(req, res)
+    #       def handle(request, response)
     #       end
     #
     #       private
@@ -184,9 +184,9 @@ module Hanami
     #
     #     class Show < Hanami::Action
     #       before { ... } # 1 do some authentication stuff
-    #       before {|req, res| @article = Article.find params[:id] } # 2
+    #       before {|request, response| @article = Article.find params[:id] } # 2
     #
-    #       def handle(req, res)
+    #       def handle(request, response)
     #       end
     #     end
     #
@@ -457,8 +457,8 @@ module Hanami
 
     # @since 0.3.2
     # @api private
-    def _requires_no_body?(res)
-      HTTP_STATUSES_WITHOUT_BODY.include?(res.status)
+    def _requires_no_body?(response)
+      HTTP_STATUSES_WITHOUT_BODY.include?(response.status)
     end
 
     # @since 2.0.0
@@ -517,10 +517,10 @@ module Hanami
 
     # @since 0.2.0
     # @api private
-    def _reference_in_rack_errors(req, exception)
-      req.env[RACK_EXCEPTION] = exception
+    def _reference_in_rack_errors(request, exception)
+      request.env[RACK_EXCEPTION] = exception
 
-      if errors = req.env[RACK_ERRORS]
+      if errors = request.env[RACK_ERRORS]
         errors.write(_dump_exception(exception))
         errors.flush
       end
@@ -534,17 +534,17 @@ module Hanami
 
     # @since 0.1.0
     # @api private
-    def _handle_exception(req, res, exception)
+    def _handle_exception(request, response, exception)
       handler = exception_handler(exception)
 
       if handler.nil?
-        _reference_in_rack_errors(req, exception)
+        _reference_in_rack_errors(request, exception)
         raise exception
       end
 
       instance_exec(
-        req,
-        res,
+        request,
+        response,
         exception,
         &_exception_handler(handler)
       )
@@ -564,15 +564,15 @@ module Hanami
 
     # @since 0.1.0
     # @api private
-    def _run_before_callbacks(req, res)
-      config.before_callbacks.run(self, req, res)
+    def _run_before_callbacks(request, response)
+      config.before_callbacks.run(self, request, response)
       nil
     end
 
     # @since 0.1.0
     # @api private
-    def _run_after_callbacks(req, res)
-      config.after_callbacks.run(self, req, res)
+    def _run_after_callbacks(request, response)
+      config.after_callbacks.run(self, request, response)
       nil
     end
 
@@ -600,16 +600,16 @@ module Hanami
     #
     #   module Books
     #     class Destroy < Hanami::Action
-    #       def handle(*, res)
+    #       def handle(*, response)
     #         # ...
-    #         res.headers.merge!(
+    #         response.headers.merge!(
     #           "Last-Modified" => "Fri, 27 Nov 2015 13:32:36 GMT",
     #           "X-Rate-Limit"  => "4000",
     #           "Content-Type"  => "application/json",
     #           "X-No-Pass"     => "true"
     #         )
     #
-    #         res.status = 204
+    #         response.status = 204
     #       end
     #
     #       private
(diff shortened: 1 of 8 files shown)


lib/hanami/action.rb
```diff
@@ -535,15 +564,15 @@ def _exception_handler(handler)
 
     # @since 0.1.0
     # @api private
-    def _run_before_callbacks(*args)
-      config.before_callbacks.run(self, *args)
+    def _run_before_callbacks(req, res)
+      config.before_callbacks.run(self, req, res)
```
timriley:
```suggestion
    def _run_before_callbacks(request, response)
      config.before_callbacks.run(self, request, response)
```

When we touch code in this repo, let's take the chance to move away from the old `req, res` shorthand names. We've deemed these names unfit for our users, so we it would be good to be consistent internally too :)

(Same suggestion for the method below this, too).

lib/hanami/action/constants.rb
```diff
@@ -67,6 +67,13 @@ class Action
     # @api private
     PATH_INFO = ::Rack::PATH_INFO
 
+    # The request content type (env key, CGI-style uppercase). Rack exposes a `CONTENT_TYPE`
+    # constant but it refers to the response header (lowercase in Rack 3), not the env key.
+    #
+    # @since x.x.x
```
timriley:
```suggestion
```

For anything that's `@api private`, I've stopped adding these `@since` tags. It's private API, and the "since" IMO falsely confers some level of stability to external readers of this code, when we should really be keeping the option to to change and refactor these as we wish.

Maybe post-release I'll suggest we do a full sweep to remove these, but for the time being I've simply stopped adding the new since tags when they're against private entries.

(This applies to all changes across this PR)

lib/hanami/action/response.rb
```diff
@@ -48,13 +48,20 @@ def self.build(status, env)
...
+        # in `spec/unit/hanami/action/response_spec.rb` if you're tempted to add one back.
+        super([], 200, headers)
         self.content_type = content_type if content_type
 
         @request = request
         @config = config
-        @charset = ::Rack::MediaType.params(content_type).fetch("charset", nil)
+        # Prefer the charset supplied by the action (computed once from config); only fall back to
+        # parsing it out of the content type when a caller constructs a Response without one.
+        @charset = charset || (::Rack::MediaType.params(content_type).fetch("charset", nil) if content_type)
```
timriley: > parsing it out of the content type when a caller constructs a response without one

I wonder if this is even possible, given the way we handle the default_charset right now?

Our `default_charset` _setting_ has a `nil` default, yes, but in `Action#initialize` we do this (where `DEFAULT_CHARSET` is `"utf-8"`):

```
default_charset = @config.default_charset || DEFAULT_CHARSET
```

AFAICT, this makes it _impossible_ to construct an action with a `default_charset` of `nil`.

So by the time we get into `#build_response`, we're always passing a real charset. Does this mean we can remove this defensive "build charset from the content type" code altogether?

In fact, if you go one level deeper and look at what builds that content type, via this attr we pass to `build_response`:

```ruby
content_type: Mime.response_content_type_with_charset(request, config)
```

You'll see that we actually do the _exact same_ "get from config but fallback to `DEFAULT_CHARSET`" approach!

```ruby
def response_content_type_with_charset(request, config)
  content_type_with_charset(
    response_content_type(request, config),
    config.default_charset || Action::DEFAULT_CHARSET # <--- look familiar?!
  )
end
```

So aside from us probably being able to drop the defensive code on this line (I suspect its vestigial from much earlier versions of hanami-controller), I feel like there's an opportunity to make all of our response charset code much more "confident", e.g. put DEFAULT_CHARSET as the default of the `default_charset` setting, and then expect a real charset at every point downstream of that.

I think this kind of refactor is probably best done as a separate PR. So I'm not expecting you to do it here (especially since this PR has so many stacked changes), but I think it'd be good to do or track nonetheless. I won't file an issue yet, but I'm keen for your thoughts on the above.

lib/hanami/action.rb
```diff
@@ -339,10 +356,21 @@ def call(env)
           session_enabled: session_enabled?,
           default_tld_length: config.default_tld_length
         )
+
+        # Only negotiate the content type when the request carries an `Accept` header; otherwise
+        # reuse the action's precomputed default.
+        content_type =
+          if request.accept_header?
```
timriley, line 359: In the real world, browsers always give us an Accept header, but there's enough programmatic API clients that don't (e.g. `Net::HTTP`'s default behaviour) that I do think this is still worth it for real world usage, especially since the up-front cost of calculating the default response content type is low.

So no changes required for this one, I just thought I'd put my thinking on the record: we should just be mindful not to contort our code for a performance bump in the edgiest of edge cases — that work should be grounded in real world usage, which I think does quite reasonably include this scenario here :)

lib/hanami/action/response.rb
```diff
@@ -71,9 +78,16 @@ def initialize(request:, config:, content_type: nil, env: {}, headers: {}, view_
       # @api public
       def body=(str)
         @length = 0
-        @body = EMPTY_BODY.dup
 
-        return if str.nil? || str == EMPTY_BODY
+        # When the body is being cleared (nil or the empty sentinel), use the frozen EMPTY_BODY
+        # constant directly. Only allocate a fresh mutable array when we actually have content to
+        # append via #write.
```
timriley: This comment doesn't feel necessary, plus it reads as very "LLM-y" and I want to keep that kind of "re-explain the mechanics that are otherwise obvious from the code below" comment out of our codebases.

timriley (approved):
Thanks @cllns, this all looks great. I've left a couple of small requests for you via inline comments, notably:

- prefer `request, response` over `req, res`, especially if we're touching such code or creating new code
- get rid of `@since` tags on `@api private` entries

Would you mind handling those before merging?

I also left some broader thoughts about how we handle the default_charset. Keen for your thoughts on those but happy to handle that separately to this PR.

lib/hanami/action.rb
cllns: Good call. Fixed in 5f59bef 

I dislike the short versions, made a note to update to the full words throughout the codebase in a separate PR.

lib/hanami/action/constants.rb
cllns: Removed all of these 386f19b

lib/hanami/action/response.rb
cllns: Great find! Definitely agree we should handle this in a more robust way in a follow-up PR. Removed the fallback for now 415d03e

lib/hanami/action/response.rb
cllns: Agreed, sorry about that. I removed some other self-explanatory comments too along with this: 1db06e8

lib/hanami/action.rb
cllns: Agreed, since this repo doesn't only serve browsers as clients, we shouldn't make assumptions about what clients will send. But I definitely agree in principle we don't want to juice the numbers for rare situations

cllns:
Thanks for the review @timriley. Addressed all the feedback, will merge soon (or feel free to do so yourself)

cllns:
Added a CHANGELOG entry, merging once CI is green

PR #522, opened 2026-06-30
Raise on unknown cache control directives, add missing ones

### LINKED ISSUE #520: The `response.cache_control` helper silently drops keys it doesn't know
## Describe the bug

I tried to set the following cache control header:

```ruby
res.cache_control :public, :immutable, max_age: 31_536_000
```

But the `immutable` directive got dropped silently, which happens here:

https://github.com/hanami/hanami-action/blob/main/lib/hanami/action/cache/directives.rb#L107

## To Reproduce

Set the `immutable` key in an action:

```ruby
def handle(req, res)
  res.cache_control :immutable
end
```

Look at the `Cache-Control` header.

## Expected behavior

Setting an unknown key should either raise or at least warn.

Aside from `immutable`, both `stale_while_revalidate: N` and `stale_if_error: N` would be useful additions as well.

We need `immutable` in four actions where we render tenant-scoped CSS and JS. The path carries a content hash, so it's safe to cache those aggressively.

## My environment

- Ruby version: 4.0.3
- OS: Ubuntu 24.04 LTS

timriley:
Previously `response.cache_control` (and the underlying Cache::Directives) silently dropped any directive it didn't recognise, so a typo or an unsupported directive vanished from the `Cache-Control` header with no feedback. Now passing an unknown directive raises an `ArgumentError`.

Also add support for several directives that were missing:

  - immutable
  - must_understand
  - stale_while_revalidate
  - stale_if_error

Update the `#cache_control` docs to describe the two directive forms, list the full supported set via @option, and point at MDN rather than the obsolete RFC 2616 link.

Fixes #520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Added
 
-- Support additional `Cache-Control` directives in `response.cache_control`: `:immutable`, `:must_understand`, `stale_while_revalidate:` and `stale_if_error:`. (@timriley)
+- Support additional `Cache-Control` directives in `response.cache_control`: `:immutable`, `:must_understand`, `stale_while_revalidate:` (requires a duration) and `stale_if_error:` (requires a duration). (@timriley in #522)
 
 ### Changed
 
-- `response.cache_control` now raises `ArgumentError` for unknown directives instead of silently dropping them. (@timriley)
+- `response.cache_control` raises `ArgumentError` for unknown directives instead of silently dropping them. (@timriley in #522)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,9 +17,12 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
     Parsed body params are symbolized as part of becoming the `Hanami::Action::Params` instance. Non-hash bodies are wrapped in `{"_" => parsed_body}`.
 - Full JRuby support. (@katafrakt in #498)
+- Support additional `Cache-Control` directives in `response.cache_control`: `:immutable`, `:must_understand`, `stale_while_revalidate:` and `stale_if_error:`. (@timriley)
```
katafrakt: Small think, but at first I did not understand why some of these have colon at the beginning and some at the end. Now after reading RFC it does make sense, but maybe for the sake of clarity it could be


```suggestion
- Support additional `Cache-Control` directives in `response.cache_control`: `:immutable`, `:must_understand` (flags), `stale_while_revalidate:` and `stale_if_error:` (user must provide a value). (@timriley)
```

? (wording could probably be improved)

katafrakt (approved):

timriley:
Thanks @katafrakt. I meant to merge this before the release, but I think I'll just go ahead and cut a patch release with this included.

CHANGELOG.md
timriley: Thanks @katafrakt — I tweaked the changelog and hopefully made things clearer :)

PR #523, opened 2026-07-04
Do not coerce session keys to strings for getters/setters

radar:
I came across a bug today where I had this code in my action's test:

```ruby
  let(:params) do
    {
      book_permalink: "book-permalink",
      chapter_permalink: "chapter-permalink",
      note_id: 1,
      text: "Updated note text",
      "rack.session" => { user_id: 1 },
    }
  end
```

But then when attempting to access the session's value in the action's code:

```ruby
request.session[:user_id] => nil
```

This is because the previous implementation was stringifying keys. There wasn't a session key stored as `"user_id"` in the session, so it returns `nil`.

This PR removes the cooercing-to-string of these values. So whatever is put in there is now maintained in the data type it was insertedin.

### DIFF
--- a/lib/hanami/action/request/session.rb
+++ b/lib/hanami/action/request/session.rb
@@ -14,6 +14,9 @@ module Hanami
 
         def_delegators \
           :@session,
+          :[],
+          :[]=,
+          :key?,
           :clear,
           :delete,
           :empty?,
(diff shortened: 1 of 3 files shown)


parndt (approved):

parndt:
Nice; I've definitely had this issue in the past.

timriley:
Thanks for this, @radar! I think this is a good change. One thing I want to do with this is have a closer poke at the ramifications for our Rack 2/3 compatibility. This string-key-ifying session wrapper was introduced in #477 as part of our Rack 3 compatibility effort, and I just want to be doubly sure that we're not introducing any regressions (the fact that our tests are happy is a good sign!).

PR #524, opened 2026-07-04
Provide `Hanami::Action::TestHelper` class

radar:
This allows for a cleaner syntax for passing through params + session data in a test:

Before:

```ruby
params = {
  book_permalink: "book-permalink",
  chapter_permalink: "chapter-permalink",
  note_id: 1,
  text: "Updated note text",
  "rack.session" => { user_id: 1 },
}
action.call(params)
```

After:

```ruby
params = {
  book_permalink: "book-permalink",
  chapter_permalink: "chapter-permalink",
  note_id: 1,
  text: "Updated note text",
}
session = { user_id: 1 }
action.call(params:, session:)
```

You can opt in to this via:

```ruby
described_class.include Hanami::Action::TestHelper
```

Or via similar "global" `before` syntax with RSpec.

Helper is opt-in, and should only be used in test environments. It is not intended for use in production code.

### DIFF
--- a/lib/hanami/action/test_helper.rb
+++ b/lib/hanami/action/test_helper.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Hanami
+  class Action
+    # Testing conveniences for calling actions directly.
+    #
+    # Prepend this module in test environments only. It allows actions to be
+    # called with a bare params hash and/or a session hash, instead of a full
+    # Rack env:
+    #
+    # @example
+    #   Hanami::Action.prepend(Hanami::Action::TestHelper)
+    #
+    #   action.call(params: {id: 1}, session: {user_id: 2})
+    #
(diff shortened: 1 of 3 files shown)


Drowze:
This is awesome! What do you think about also accepting a `headers` keyword? E.g.:

```ruby
# before
action.call({ id: 42, "HTTP_AUTHORIZATION" => "Bearer foo", "HTTP_X_CUSTOM" => "xyz" })

# after
action.call(params: { id: 42 }, headers: { authorization: "Bearer foo", "x-custom": "xyz" })
```

timriley:
Thanks for putting this together, @radar! I agree about the value of a convenience layer for testing. I'm just exploring a possible alternative API we could try and will post some thoughts here again soon.

(@Drowze — agree that `headers:` could be valuable too!)

Issue #529, opened 2026-09-08
Hanami actions cannot stream a response body


aaronmallen:
I have been building an application on Hanami 3.0 that has to serve two kinds of large response: a plain text document assembled from tens of thousands of database rows, and file downloads that can run to several megabytes each. On every other Rack framework I have used, both of those go out incrementally. On Hanami they cannot. Every response is buffered in full before a single byte reaches the client.

I do not think this is a bug in my code. I think `Hanami::Action` has no streaming path at all, by construction, and I would like to know whether the team sees that as a gap worth closing.

Everything below is against `hanami-action` 3.0.1 and `rack` 3.2.7. I checked `main` before writing this and the relevant code is unchanged there.

## The mechanism

`Hanami::Action::Response` inherits from `Rack::Response`, so the plumbing for a lazy body is already underneath it. The one door into that body is `Response#body=`, at `lib/hanami/action/response.rb:77`:

```ruby
def body=(str)
  @length = 0

  if str.nil? || str == EMPTY_BODY
    @body = EMPTY_BODY
    return
  end

  @body = []

  if str.is_a?(::Rack::Files::BaseIterator)
    @body = str
    buffered_body!          # reads the whole thing, to set content-length
  else
    write(str)              # appends to the buffer
  end
end
```

There are three paths and all three end in a buffer. A String is appended. A `Rack::Files::BaseIterator` is drained by `buffered_body!` into an array so that a content-length can be set. Anything else, an Enumerator, an object that responds to `each`, a `Rack::BodyProxy`, is handed to `write`, which calls `to_str` on it and fails. There is no fourth path, and no other writer for `@body` on the public surface.

`Response#write` has the same effect by another route. It appends to `@body` and increments `@length`. Calling it once per batch, which is the obvious thing to reach for, buys nothing: the batches accumulate and the whole body still goes out in one piece at the end of the action.

`send_file` is not an escape hatch either. It goes through `_send_file` at `response.rb:477`, which does `self.body = send_file_response[RESPONSE_BODY]`, which is the `Rack::Files::BaseIterator` branch above. So Hanami's own file sending reads the entire file into memory before responding. On a large file that is the exact behaviour I am trying to avoid, and it is the framework doing it rather than my code.

The consequence is a hard ceiling on what a Hanami action can serve. Memory per request scales with response size, and time to first byte scales with total response time. For an endpoint that assembles a document out of a large result set, that is the difference between a client seeing output immediately and waiting for the whole thing.

[...]

PR #530, opened 2026-09-13
Add response streaming

timriley:
I was inspired to implement an API matching @katafrakt's proposal in @aaronmallen's fantastic report in #529. Let me know what you think!

Here we introduce `Response#stream` as a way to set a lazy response body, which Rack sends to the client one chunk at a time:

```ruby
response.stream do |out|
  # Imagine `serialize` here turning our record batches into into strings
  records.each_slice(500) { |batch| out << serialize(batch) }
end
```

Behind the scenes, this is initialising a new `Hanami::Action::Stream` and assigning it to the body. You can do this, too, if you want to create the stream separately:

```ruby
response.body = Stream.new { |out|
  records.each_slice(500) { |batch| out << serialize(batch) }
}
```

Streams can receive either an enumerable of string chunks (`response.stream(chunks)` or `Stream.new(chunks)`) or use the block-based writing, as above.

As part of this, tighten up related behavior in `Response`:

- `#body=` raises for an object that responds to `#each`, such as an Array or an Enumerator. It used to send the `#inspect` output of the object to the client. Responding to `#each` isn't enough for us to be confident an object is a fully valid Rack body, and since we hand the body to the server untouched, anything wrong with it only surfaces once the headers have gone out. Now that we have `Stream`, that is what you can use to supply an enumerable body we can trust.
- `#write` raises when the body is a stream, which would otherwise buffer the whole stream and undo it.
- `#body=` closes the body that it replaces, which is required by Rack. Without this, if a response had a stream body set, but then replaced, the initial stream would keep its source open.

See also the second commit in this PR, where we use `Stream` to wrap up file downloads, which means we no longer buffer the entire thing into memory before sending it. Phew!


lib/hanami/action/response.rb
```diff
@@ -70,28 +70,116 @@ def initialize(request:, config:, content_type: nil, charset: nil, env: {}, head
...
+          headers.delete(Action::CONTENT_LENGTH)
 
-        if str.is_a?(::Rack::Files::BaseIterator)
-          @body = str
-          buffered_body! # Ensure appropriate content-length is set
+          # Rack 2 sets the content-length header as a body is written, rather than at `finish`, so
+          # set it here. Rack 3 sets it from `@length` in `finish`, so this can go when we drop Rack
+          # 2 support.
+          headers[Action::CONTENT_LENGTH] = @length.to_s if @length
+        when ::Rack::Files::BaseIterator
```
alassek, line 116: My initial reaction to this was: is an explicit type check actually necessary here, or could this be duck-typed somehow?

Thinking through why someone would want that:
- you have a custom serializer class
- you want to control the streaming process in a central place
- you want to delegate the decision whether to stream to something else

Okay, but what if you just wrote something like `#to_stream` on your objects. Would it actually be better to write

```ruby
response.body = thing
```

instead of

```ruby
response.body = thing.to_stream
```

actually no, I think being explicit that you are returning a stream in the action is a benefit. The `Stream` interface is simple to use, and the `response.stream` interface is perfectly easy if you don't want your objects to know about this.

In fact, it appears that `response.stream` accepts any enumerable object as an argument, so while not front-and-center, duck-typing is accounted for.

I think my intuition was wrong in this case, this is a good solution.

CHANGELOG.md
```diff
@@ -10,15 +10,24 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 ### Added
 
 - Support for callbacks using ordinary objects that respond to `call`; previously, only `#to_proc` objects were accepted. (@ziggythehamster in #528, via hanami/hanami-utils#420)
+- `Hanami::Action::Stream`, a lazy response body. Assign a stream to `Response#body=` to send a response in chunks instead of buffering it.
+- `Response#stream`, a convenience for making a `Stream` and assigning it to the body: `response.stream { |out| out << chunk }`. Give it an enumerable of String chunks or a block, just like `Stream` itself.
```
alassek, line 14: Did you mean "a convenience method"?

lib/hanami/action/stream.rb
```diff
@@ -0,0 +1,182 @@
...
+        return to_enum(:each) if callback.nil?
+
+        if @closed
+          raise StreamError, "This stream is closed. A response body can be sent only once."
+        end
+
+        writer = Writer.new(callback)
+
+        if @block
+          @block.call(writer)
```
alassek, line 133: This class is a little unusual in the way it's interleaving multiple blocks together here. Would it be beneficial to give this one a more intention-revealing name?

alassek (approved):
I like the simple design of `Stream` and the `response.stream` interface, I think this does what we need well.

lib/hanami/action/stream.rb
```diff
@@ -0,0 +1,182 @@
...
+      #
+      # @return [Integer, nil]
+      #
+      # @since 3.1.0
+      # @api public
+      attr_reader :length
+
+      # Returns a new stream.
+      #
+      # Give either an enumerable of chunks or a block, but not both.
```
citizen428, line 88: Given the constraint stated here and the XOR check in `initialize`, would it make sense to provide "factory methods" like `Stream.from_chunks(chunks, length: nil)` and `Stream.from(length: nil, &block)` [1]? One could even throw in a `private_class_method :new`. To me, this feels preferable to a fallible constructor.

[1] Naming is hard. You get the idea.

aaronmallen (approved):
LGTM :shipit:

